PR #15, opened 2019-04-28
Added tile selector window

2767mr:
The old tile selector that appears when pressing space still works.


webapp/src/app/shared/phaser/tilemap/tile-drawer.ts
```diff
@@ -358,6 +366,34 @@ export class TileDrawer extends Phaser.Plugin {
 		});
 	}
 	
+	private renderPreviewById() {
```
Vegita2: This method contains some duplicated code. Maybe modify the renderPreview function to cover both use cases or just extract the duplicated code

PR #22, opened 2019-05-02
Added common directory to avoid code duplication

2767mr:
The common directory contains shared code between backend and webapp. It is intended for fs operations


common/src/require.ts
```diff
@@ -0,0 +1,17 @@
+let result: NodeRequireFunction | (() => null);
+
+if (!!module['require']) {
```
Vegita2: I don't understand this line. 
if `module['require']` exists, then the platform is not supported, but inside the if block `module['require']` is assigned to the result, so it should exist?

Vegita2:
The changes look good. Only thing that's missing is an update to the readme which contains steps for the common module and that it's important to do that first

PR #23, opened 2019-05-03
Added PersonExpression Widget

ac2pic:


webapp/src/app/shared/widgets/person-expression-widget/person-expression-widget.component.html
```diff
@@ -0,0 +1,6 @@
+<app-character-widget [custom]="settings[key]" key="person"></app-character-widget>
+
+<div fxLayout="row" class="property-container" [matTooltip]="attribute?.description">
+	<span class="property-name">expression:</span>
```
Vegita2: use a label instead of span. When clicking on a label the cursor is automatically placed inside the input. You can see an example how to do that in the string widget.
I forgot to mention it in the other pr

ac2pic:
Fixed

PR #27, opened 2019-05-09
Create Empty Map

ac2pic:


ac2pic:
This is not the latest code.

ac2pic:
I updated it to use a dialogue.

PR #28, opened 2019-05-10
Fixed tileselector on new layer

### LINKED ISSUE #19: Selecting a Newly Created Layer breaks TileSelector

2767mr:
Fixes #19


ac2pic:
Unless we can control what is entered into the tilesetName, this will break when someone puts in an invalid tileset path.

2767mr:
It will not break but display a "unable-to-load" icon

ac2pic:
Great.

PR #37, opened 2019-06-08
Updated dependencies and build scripts for all modules

2767mr:
* Removed unused asset copy from backend
* Removed unused test script from backend
* Simplified watch and run for backend
* Updated dependencies for backend, common and webapp
* Removed deprecated tslint option from backend
* Added start script for common
* Fixed errors in webapp caused by angular upgrade


ac2pic:
Cleanup is 30 commits behind master.

2767mr:
That does not matter as these changes do not conflict with the changes on the master branch.

webapp/src/main.ts
```diff
@@ -5,6 +5,8 @@ import {AppModule} from './app/app.module';
 import {environment} from './environments/environment';
 import {Globals} from './app/shared/globals';
 
+import 'hammerjs';
```
Vegita2: why is hammerjs imported?

2767mr: Mostly because Angular Material wants to have it and throws a warning otherwise (see [Angular Material docs](https://material.angular.io/guide/getting-started#step-5-gesture-support) ). It is just a library to improve touch support and looking at the source code it is so small that it would not hurt at all to include it. I am not sure it it is actually used by the components we are using but it's a possibility.

2767mr: If the question was: Why was it specifically imported here? This ensures that the treeshaking doesn't remove the dependency in the build.

Vegita2: Alright, makes sense.

PR #43, opened 2019-06-19
Fixes for zoom

### LINKED ISSUE #41: zoom doesn't work
zoom isn't working. Phaser changed their api 

`input.mouse.mouseWheelCallback` → `input.mouseWheel.callback`

2767mr:
Contains a workaround for missing typings that should be removed once phaser-ce fixes it.
Fixes #41


Vegita2:
Looks good. We should fix phaser-ce version so something like that doesn't happen again

PR #61, opened 2019-07-19
Add float support for Numbers

ac2pic:


webapp/src/app/shared/widgets/abstract-widget.ts
```diff
@@ -33,10 +33,6 @@ export abstract class AbstractWidget implements OnInit, OnChanges {
 		}
 	}
 	
-	toInt(val: string): number {
```
Vegita2: this function is still needed in vec2 widget

webapp/src/app/shared/widgets/number-widget/number-widget.component.ts
```diff
@@ -1,14 +1,38 @@
...
-export class NumberWidgetComponent extends AbstractWidget {
-	constructor() {
-		super();
-	}
+export class NumberWidgetComponent extends AbstractWidget implements OnInit {
+	@ViewChild('numberInput', {static: true}) input!: ElementRef;
+
+	private lastValidValue = '0';
+	
+	setSettings(key: string, event: any) {
```
Vegita2: don't use any. Use the actual type

webapp/src/app/shared/widgets/number-widget/number-widget.component.ts
```diff
@@ -1,14 +1,38 @@
...
+		
+
+		let value = event.target.value;
+		
+		if (event.data) {
+			if (event.data.length > value.length) {
```
Vegita2: what is that even doing? google says event.data returns a string with the inserted characters. That means it should never be > value.length because that one contains all characters.

Another Point is setting it to the last valid value. I would prefer it to not set anything in this case and just show a warning/error

webapp/src/app/shared/widgets/number-widget/number-widget.component.ts
```diff
@@ -1,14 +1,38 @@
...
+			}
+		}
+
+		// Number converts "" to 0
+		// while parseFloat converts "" to NaN
+		const num = Number(value);
 
+		this.lastValidValue = num.toString();
+
+		this.input.nativeElement.value = num.toString();
```
Vegita2: There is no need to get the input element as a ViewChild. ngModel on the html element handles the value and it should not be overriden by directly accessing the native element.

webapp/src/app/shared/widgets/vec2-widget/vec2-widget.component.html
```diff
@@ -6,14 +6,12 @@
 	       [step]="step"
 	       [min]="minSize.x"
 	       [disabled]="!enableX"
-	       [ngModel]="settings[key].x"
-	       (input)="setVal('x', toInt($event.target.value))">
+	       [(ngModel)]="settings[key].x">
```
Vegita2: vec2 widget still needs the toInt function

webapp/src/app/shared/widgets/number-widget/number-widget.component.ts
ac2pic: What is that even doing?
if  event.data === null, then the user backspaced.
if value.length === 0 && event.data.length === 0, then the value was reset by the input property because the input was invalid.

I felt that it would be frustrating to accidentally type in something invalid for it to decide to reset itself.

webapp/src/app/shared/widgets/number-widget/number-widget.component.ts
ac2pic: There is a need when I need to control the input.

ac2pic: angular does not behave how I want it to if the same value is set twice.

webapp/src/app/shared/widgets/vec2-widget/vec2-widget.component.html
ac2pic: Will fix.

webapp/src/app/shared/widgets/number-widget/number-widget.component.ts
```diff
@@ -1,14 +1,24 @@
...
 	templateUrl: './number-widget.component.html',
 	styleUrls: ['./number-widget.component.scss', '../widget.scss']
 })
-export class NumberWidgetComponent extends AbstractWidget {
-	constructor() {
-		super();
-	}
+export class NumberWidgetComponent extends AbstractWidget implements OnInit {
+
+	setSettings(key: string, value: string) {
```
Vegita2: this function name is easily mistaken for the other setSetting function. It would make more sense to give it a completely different name or just override the setSetting method

PR #62, opened 2019-07-22
Add a map selector

2767mr:


2767mr:
> there is no straightforward way of closing it.
You can close the sidenav by clicking on the side (where the rest of the editor is) as is standard in web applications. I will also add a close button as requested.

> a checkbox to toggle a flat view, so one can filter and directly select the map without clicking through folders
This would be easy to implement but would be hard to use since there are a lot of maps and would also cause performance issues. It would only be usable with a filter of at least 2 characters.

webapp/src/app/components/load-map/virtualMapNode.model.ts
```diff
@@ -0,0 +1,48 @@
...
+        return this.original.path;
+    }
+
+    public get children(): VirtualMapNode[] | undefined {
+        if (!this.original.children) {
+            return undefined;
+        }
+
+        return this.original.children
+            .filter((n): boolean => n.displayed)
```
Vegita2: is "boolean" really needed? `.filter(n => n.displayed)` should work

webapp/src/styles.scss
```diff
@@ -93,6 +97,11 @@ button {
 	}
 }
 
+
+.dark-scrollbar .mat-drawer-inner-container {
```
Vegita2: is this still used somewhere?

webapp/src/styles.scss
```diff
@@ -102,3 +111,6 @@ button {
 	}
 }
 
+input[type="file"] {
```
Vegita2: i wouldn't add it to global styles. Someday someone wants to add another file input and gets confused as to why it's not there

webapp/src/app/components/load-map/load-map.component.scss
```diff
@@ -0,0 +1,38 @@
...
+	list-style-type: none;
+}
+
+.mapTree li > .mat-button {
+	text-transform: none;
+	width: 100%;
+	min-height: 48px;
+}
+
+.mapTree li > .mat-button /deep/ .mat-button-wrapper {
```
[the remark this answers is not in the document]
2767mr: Removing this would center the text and icon of every button which is not desirable.
image

webapp/src/app/components/load-map/virtualMapNode.model.ts
2767mr: I could have sworn the linter requires these but it seems I was wrong. Fixed.

webapp/src/styles.scss
2767mr: I changed it to make it a bit more obvious although we probably won't add visible file inputs as they are not supported by Angular Material

webapp/src/app/components/load-map/load-map.component.scss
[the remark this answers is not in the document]
Vegita2: adding a class to the button with `text-align: left;` fixes it

webapp/src/app/components/load-map/load-map.component.html
```diff
@@ -0,0 +1,50 @@
...
+			</mat-form-field>
+		</div>
+		<input #fileUpload type="file" class="hidden" (change)="loadMap($event)" accept=".json"/>
+	</div>
+	<div fxFlex class="map-list dark-scrollbar">
+		<mat-tree [dataSource]="mapsSource" [treeControl]="treeControl" class="mapTree">
+			<mat-tree-node *matTreeNodeDef="let node">
+				<li class="mat-tree-node">
+					<button mat-button class="mapTree-button" (click)="node.path ? load(node.path) : fileUpload.click()">
+						<mat-icon class="icon mat-icon">edit</mat-icon>
```
Vegita2: is that mat-icon class necessary?

PR #66, opened 2019-08-12
Height generator + autotiles

Vegita2:
Adds Height map generation with support for autotiles (and standalone autotiles).

Autotiles are automatically used if a tile from the autotile sheet is selected. When holding shift autotiles are ignored.


Vegita2:
- I'm not really satisfied with the filenames and naming of some variables
- needs some refactoring for shared constants
- configurations for autotiles / cliffs should be extracted to json to support custom tilesets

Not really sure when i get to fix that, but i think these additions make the editor finally usable to create new maps.

PR #80, opened 2019-08-22
Entity fixes

### LINKED ISSUE #30: Can't add or change NPC if characterName is invalid

### LINKED ISSUE #71: npcStates menu
When creating an NPC entity, you can edit it's settings, like usual.
image
But when you click on "npcStates", you get greeted with a menu, to change states of an NPC, I presume. The issue with this very menu is, that it is unclosable.
image
Pressing either "Cancel" or "Save" will not do anything, and neither will clicking outside of the menu.

### LINKED ISSUE #75: Person autofill
When having to select a "person" in an event, or an NPC, there's an autofill suggestion that pops up.
image
This does not work.
When you click it, it appears in the text field, and when saving and re-opening the screen, the "person" field is empty again. You'll have to manually type it in.

Vegita2:
Fixes some Entity related bugs.

Fixes #71
Fixes #75
Fixes #30


webapp/src/app/shared/phaser/entities/registry/item-destruct.ts
```diff
@@ -48,7 +48,8 @@ export class ItemDestruct extends CCEntity {
 		return undefined;
 	}
 	
-	protected setupType(settings: any) {
+	protected async setupType(settings: any) {
+		// TODO: use await
```
2767mr: I assume this was left intentionally

webapp/src/app/shared/phaser/entities/registry/npc.ts
```diff
@@ -116,6 +121,7 @@ export class NPC extends CCEntity {
 		let x = charSettings.x || 0;
 		let y = charSettings.y || 0;
 		if (charSettings.animSheet) {
+			// noinspection SuspiciousTypeOfGuard
```
2767mr: Why was this added? All builds and linters pass without this line. If  `charSettings.animSheet` can indeed be a string then the types are wrong.

webapp/src/app/shared/phaser/entities/registry/prop.ts
```diff
@@ -89,11 +89,12 @@ export class Prop extends CCEntity {
 		return undefined;
 	}
 	
-	protected setupType(settings: any) {
+	protected async setupType(settings: any) {
 		if (!settings.propType) {
 			console.warn('prop without prop type');
 			return this.generateNoImageType();
 		}
+		// TODO: use await
```
2767mr: I assume this was left intentionally

webapp/src/app/shared/phaser/entities/registry/scalable-prop.ts
```diff
@@ -80,11 +80,12 @@ export class ScalableProp extends CCEntity {
 		return this.scaleSettings;
 	}
 	
-	protected setupType(settings: any) {
+	protected async setupType(settings: any) {
 		if (!settings.propConfig) {
 			console.warn('scalable prop without prop config');
 			return this.generateNoImageType();
 		}
+		// TODO: use await
```
2767mr: I assume this was left intentionally

webapp/src/app/shared/widgets/event-widget/event-registry/ShowChoice.ts
```diff
@@ -73,7 +73,8 @@ export class ShowChoice extends AbstractEvent<ShowChoiceData> {
 			if (!child.events) {
 				console.error('wtf', this);
 			}
-			out[index as keyof ShowChoiceData] = child.events.map(v => v.export());
+			// @ts-ignore
```
2767mr: Why was this changed, especially in an PR that modifies entity? Also, before adding exceptions like `@ts-ignore` the type should be changed to fit the demand.

2767mr:
Contains fixes for #30

webapp/src/app/shared/phaser/entities/registry/npc.ts
Vegita2: Linters pass, but typescript itself complains after updating. 
Problem is the type only matches for the number props not the other and i couldn't figure out how to fix that and still have a somewhat readable code

webapp/src/app/shared/widgets/event-widget/event-registry/ShowChoice.ts
Vegita2: Linters pass, but typescript itself complains after updating. 
Problem is the type only matches for the number props not the other and i couldn't figure out how to fix that and still have a somewhat readable code

2767mr: Defining the indexer `[index: number]: AbstractEvent<any>[];` on `ShowChoiceData` should be able to solve that problem.

webapp/src/app/shared/phaser/entities/registry/npc.ts
Vegita2: The types are wrong in this case. The input can be a string and that's why it's catched here and filled with the correct object.

Input should probably have another type to reflect this, but the whole npc entity misses a few cases. We should first understand how the game handles all the different cases and then see if this solution still makes sense before refactoring it

webapp/src/app/shared/widgets/event-widget/event-registry/ShowChoice.ts
Vegita2: That's a good idea, didn't thought of that

webapp/src/app/shared/phaser/entities/registry/npc.ts
2767mr: In that case, adding a `//TODO` should be enough for this PR.

PR #94, opened 2019-09-02
improved default events

### LINKED ISSUE #78: add properties to default event
default events currently only show a textarea with the raw json. Better would be to get the properties from a json and display them instead

Vegita2:
Depends on #80

Improves default event by adding a database of all events with the associated attributes. 
The default event reads it and shows all the editable properties in the event editor instead of a single json.

fixes #78

image


keanuplayz:
I just tested it and all works fine. @ac2pic and @2767mr saw it.
You can add all events manually, and edit them without issues.

PR #97, opened 2019-09-04
Improved default entity

### LINKED ISSUE #79: add properties to default entity
similar to #78 

Currently default entity shows only the properties defined in that instance. It would be better to get that info from a json to show all available properties

Vegita2:
Same as #94 but for entities.

Additionally filterable list is refactored into reusable component (used by add entity and add event)

fixes #79

image


webapp/src/app/components/editor/editor.component.ts
```diff
@@ -16,10 +14,12 @@ export class EditorComponent {
 	
 	@ViewChild('sidenavLoadMap', {static: true})
 	sidenavLoadMap!: MatSidenav;
-
-	constructor() {
+	
+	constructor(
+		addEntityMenu: AddEntityMenuService
```
2767mr: Unused dependency injection

webapp/src/styles.scss
```diff
@@ -95,6 +95,9 @@ button {
 	&::-webkit-scrollbar-thumb {
 		background-color: mat-color($accent, 600);
 	}
+
+	//noinspection CssInvalidPropertyValue
+	overflow: overlay;
```
2767mr: Why is this needed? What does it do? Why does it need a special directive to disable the linter? Add comments to code if needed.

webapp/src/app/components/editor/editor.component.ts
Vegita2: it needs to be injected somewhere so an instance is created and the service can initialize itself. 
I move the code to an init function and call that to make it more obvious

webapp/src/styles.scss
Vegita2: the comment is webstorm specific and should be ignored by the linter. 
`overlay` is a non standard value only used by webkit based browser that makes the scrollbar not take any space, so the content doesn't get resized when the scrollbar is visible.

I'll add a comment which links to a definition

Vegita2:
I cannot reproduce the issue with the context menu in chrome and firefox. Can you give me more information about it?

PR #117, opened 2019-09-20
Open in CC.

keanuplayz:
Adds open in CC button by @2767mr


2767mr:
Use the [Map Watcher](https://github.com/CCDirectLink/map-watch) mod instead.

PR #123, opened 2019-12-28
Mod support 

ac2pic:


ac2pic:
There is a major performance issue when running through electron. Have to fix that.

Vegita2:
can you explain a bit what this pr does?

ac2pic:
Settings > Map Context  allows you to select a mod and view it's files. 
It also automatically handles resource resolution similar to how CCLoader would.

ac2pic:
What I still need to do is allow the ResourceResolver cache to be cleared.

Vegita2:
When i open the settings the only option available for map context is `BASE` and i cannot load any maps from this mod https://github.com/2hh8899/ArcaneLab

webapp/src/app/components/dialogs/settings/settings.component.ts
```diff
@@ -56,23 +74,33 @@ export class SettingsComponent implements OnInit {
...
-			duration: 6000
-		});
-		
-		ref.onAction().subscribe(() => this.electron.relaunch());
 	}
 	
+
+	selectMod(index: number) {
+		const mod = this.mods[index];
+		Globals.globalEventsService.changeMapContext.next(mod);
```
2767mr: Use dependency injection instead of `Globals`

webapp/src/app/components/dialogs/settings/settings.component.ts
```diff
@@ -56,23 +74,33 @@ export class SettingsComponent implements OnInit {
...
 		this.close();
-		const ref = this.snackBar.open('Changing the path requires to restart the editor', 'Restart', {
-			duration: 6000
-		});
-		
-		ref.onAction().subscribe(() => this.electron.relaunch());
 	}
 	
+
+	selectMod(index: number) {
```
2767mr: Prefer data entry over index

webapp/src/app/components/dialogs/settings/settings.component.ts
```diff
@@ -1,34 +1,52 @@
...
 })
 export class SettingsComponent implements OnInit {
 	
+	isElectron = Globals.isElectron;
 	folderFormControl = new FormControl();
+	mapContextFormControl = new FormControl();
 	icon = 'help_outline';
 	iconCss = 'icon-undefined';
-	
+	mods: any[] = [{name: 'BASE'}];
```
2767mr: Do not use `any`

webapp/src/app/components/load-map/load-map.component.ts
```diff
@@ -32,12 +32,24 @@ export class LoadMapComponent {
...
 	constructor(
 		private mapLoader: MapLoaderService,
 		private http: HttpClientService,
+		private eventService: GlobalEventsService
 	) {
 		this.mapsSource.data = [];
 		this.refresh();
+
+		this.eventService.changeMapContext.subscribe(({name, path: modPath}: any) => {
+			if (name === 'BASE') {
```
2767mr: It would make more sense to handle this in `this.http.getMaps`

webapp/src/app/components/dialogs/settings/settings.component.html
```diff
@@ -13,12 +13,22 @@
...
-					<button mat-stroked-button (click)="check()">
+					<button mat-stroked-button (click)="check() && initMods()">
 						Check
 						<mat-icon [ngClass]="iconCss" class="icon-base">{{icon}}</mat-icon>
 					</button>
 				</div>
 			</form>
+			<form fxLayout="row" class="option-row">
+				<mat-form-field fxFlex="auto">
+					<mat-label>Map Context</mat-label>
```
2767mr: `Map Context` can be a bit confusing because it isn't directly related to modding. Maybe use `Mod`, `Mod map filter` or similar

backend/src/server.ts
```diff
@@ -35,6 +40,25 @@ app.post('/api/saveFile', async (req, res) => {
 	}
 });
 
+
+app.get('/api/mods/assets/path', async (_, res) => {
```
2767mr: I'd prefer `'/api/mods'` because it's shorter and already makes sense (`GET` me the `mods`)

webapp/src/app/services/modloader.service.ts
```diff
@@ -0,0 +1,63 @@
+import { Injectable } from '@angular/core';
+import {Globals} from '../shared/globals';
+import {GenericModLoader, Mod} from '@ac2pic/modloader';
+import {Observable} from 'rxjs';
+import {HttpClient} from '@angular/common/http';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class ModloaderService {
```
2767mr: Is this service really needed? It would make a lot more sense to hide this code in http and move all relevant logic to `common` while consolidating with the code from `backend`

webapp/src/app/services/modloader.service.ts
```diff
@@ -0,0 +1,63 @@
...
+    ) { 
+        if (Globals.isElectron) {
+            // @ts-ignore
+            this.modloader = new (require('@ac2pic/modloader'));
+        } else {
+            this.modloader = null;
+        }
+    }
+
+    loadJson(relativePath: string): Promise<any> {
```
2767mr: Do not use `any`

webapp/src/app/services/modloader.service.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    getResourcePath(relativePath: string): Observable<string> {
+        if (this.modloader) {
+            const resourcePath = this.modloader.getResourcePath(relativePath);
+            return new Observable<string>(sub => sub.next(resourcePath));
+        }
+        return this.http.get<string>(Globals.URL + 'api/resource/path/?path=' + encodeURI(relativePath));
+    }
+
+    getAllModsAssetsPath(): Observable<{name: string, path: string}[]> {
```
2767mr: `Observable<Map<string, string>>` would be a better type. It would be even more expressive if `Map<string, string>` was a named type.

webapp/src/app/shared/globals.ts
```diff
@@ -26,4 +27,5 @@ export class Globals {
 	static phaserEventsService: PhaserEventsService;
 	static autotileService: AutotileService;
 	static entityRegistry: EntityRegistryService;
+	static modloaderService: ModloaderService;
```
2767mr: Do not add more services to global if possible. As the `//TODO` suggests we are trying to remove them.

webapp/src/app/shared/map-loader.service.ts
```diff
@@ -62,7 +71,7 @@ export class MapLoaderService {
 	}
 	
 	loadMapByName(name: string) {
-		const path = PathResolver.convertToPath(BasePath.MAPS, name, FileExtension.JSON);
+		const path = PathResolver.convertToPath(this.pathOverride + BasePath.MAPS, name, FileExtension.JSON);
```
2767mr: Consider adding this as argument too `PathResolver.convertToPath`

webapp/src/app/shared/path-resolver.ts
```diff
@@ -15,7 +15,7 @@ export class PathResolver {
 	 * @param name The name of the asset in relation to base using '.' as separator.
 	 * @param extension
 	 */
-	public static convertToPath(base: BasePath, name: string, extension: FileExtension) {
+	public static convertToPath(base: BasePath | string, name: string, extension: FileExtension) {
```
2767mr: Consider adding the map context as argument

webapp/src/app/shared/phaser/helper.ts
```diff
@@ -65,32 +65,30 @@ export class Helper {
...
 		return JSON.parse(JSON.stringify(obj));
 	}
-	
+		
+
 	public static getJson(key: string, callback: (json: any) => void) {
 		const scene = Globals.scene;
-		
-		// get json from cache
+
```
2767mr: Cache will return old data. Consider including the map context in the scene cache key.

webapp/src/app/shared/phaser/helper.ts
```diff
@@ -102,10 +100,12 @@ export class Helper {
 		}
 		
 		return new Promise(res => {
-			scene.load.image(key, Globals.URL + key);
-			scene.load.once('complete', () => res(true));
-			scene.load.once('loaderror', () => res(false));
-			scene.load.start();
+			Globals.modloaderService.getResourcePath(key).subscribe((truePath) => {
```
2767mr: Cache will return old data

2767mr: Use `await` for clean code

webapp/src/app/components/dialogs/settings/settings.component.ts
```diff
@@ -56,23 +74,33 @@ export class SettingsComponent implements OnInit {
...
 			});
+			return false;
 		}
 	}
 	
 	save() {
-		this.electron.saveAssetsPath(this.folderFormControl.value);
+		if (Globals.isElectron) {
+			this.electron.saveAssetsPath(this.folderFormControl.value);
+			const ref = this.snackBar.open('Changing the path requires to restart the editor', 'Restart', {
```
2767mr: Only display this if path was actually changed

ac2pic: That's a general issue with the map editor.

webapp/src/app/shared/phaser/helper.ts
ac2pic: There would no longer be a point to getJsonPromise. I don't even see the reason it's still there.

webapp/src/app/shared/phaser/helper.ts
ac2pic: Intended. The map editor never clears the cache so this is never an issue.

webapp/src/app/shared/phaser/helper.ts
ac2pic: Same as above.

webapp/src/app/shared/path-resolver.ts
ac2pic: Good point.

ac2pic:
Yeah. I probably should just remove it from settings.

webapp/src/app/shared/phaser/helper.ts
ac2pic: I was looking at the wrong code.

webapp/src/app/shared/phaser/helper.ts
```diff
@@ -100,9 +109,13 @@ export class Helper {
 		if (scene.textures.exists(key)) {
 			return true;
 		}
-		
+		const truePath = await Helper.modloaderService.getResourcePath(key);
 		return new Promise(res => {
-			scene.load.image(key, Globals.URL + key);
+			if (Globals.isElectron) {
```
Vegita2: `Helper.modloaderService.getResourcePath(key);` should handle differences between electron and web build

webapp/package.json
```diff
@@ -66,6 +67,7 @@
 		"@angular/platform-browser": "^8.2.3",
 		"@angular/platform-browser-dynamic": "^8.2.3",
 		"@angular/router": "^8.2.3",
+		"@types/node": "^13.1.1",
```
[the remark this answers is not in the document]
ac2pic: Nope. I just installed the latest.

webapp/src/app/components/dialogs/map-context/map-context.component.html
```diff
@@ -0,0 +1,13 @@
+<cc-overlay-panel title="Map Filter">
+	<ng-container ngProjectAs="content">
+		<mat-radio-group class="radio-group" [(ngModel)]="selected.name">
+			<mat-radio-button class="radio-button" *ngFor="let mod of mods" [value]="mod.name" (change)="onSelect($event, mod)">
```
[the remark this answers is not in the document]
ac2pic: Okay.

webapp/src/app/components/dialogs/map-context/map-context.component.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  close() {
+    this.ref.close();
+  }
+
+  save() {
+    this.events.changeMapContext.next(this.selected);
+    this.close();
+  }
+}
```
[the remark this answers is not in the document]
ac2pic: Yeah. It was auto generated and I was too lazy to fix it right away.

webapp/src/app/services/modloader.service.ts
```diff
@@ -0,0 +1,51 @@
+import { Injectable } from '@angular/core';
```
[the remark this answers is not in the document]
ac2pic: I refuse.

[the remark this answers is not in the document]
ac2pic: I will do it with the knowledge that the name is going to be changed.

webapp/src/app/shared/phaser/helper.ts
```diff
@@ -62,35 +69,37 @@ export class Helper {
 	}
 	
 	/** copies obj via JSON.parse(JSON.stringify(obj)); */
-	public static copy(obj: any) {
+	public static copy<T>(obj: T): T {
```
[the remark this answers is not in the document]
ac2pic: Maybe. I don't know where.

webapp/src/app/shared/phaser/helper.ts
```diff
@@ -62,35 +69,37 @@ export class Helper {
...
-		// load json
-		scene.load.json(key, Globals.URL + key + '.json');
-		scene.load.once('complete', () => {
-			return callback(scene.cache.json.get(key));
+		Helper.modloaderService.loadJson(key + '.json').then((data: T) => {
+			scene.cache.json.add(key, data);
+			callback(data);
+		}).catch(() => {
+			// whenever there is an issue for any reason
+			scene.cache.json.add(key, null);
```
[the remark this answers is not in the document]
ac2pic: Okay

webapp/src/app/services/modloader.service.ts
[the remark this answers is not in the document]
Vegita2: just change the name in this pr

PR #143, opened 2020-03-15
Combine folders with only one subfolder into a singe node in the load-map dialogue

### LINKED ISSUE #111: Map folder list is messy.
image
The map folder list does not stay in alphabetical order, and folders jump positions when you open a map in them. I recommend keeping them alphabetical at all times.

2767mr:
image

Fixes #111

### DIFF
--- a/webapp/src/app/components/load-map/virtualMapNode.model.ts
+++ b/webapp/src/app/components/load-map/virtualMapNode.model.ts
@@ -18,16 +18,11 @@ export class VirtualMapNode {
     }
 
     public get names(): string[] {
-        const children = this.realChildren;
-        if (!children
-            || children.length !== 1 
-            || !children[0].isDirectory
-            || this.isRoot
-        ) {
+        if (!this.containsSingleDirectory) {
             return [this.original.name];
         }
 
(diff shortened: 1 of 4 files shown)


webapp/src/app/components/load-map/virtualMapNode.model.ts
```diff
@@ -17,22 +17,33 @@ export class VirtualMapNode {
...
 
     public get children(): VirtualMapNode[] | undefined {
-        if (!this.original.children) {
-            return undefined;
+        const result = this.realChildren;
+        if (result 
+            && result.length === 1 
+            && result[0].isDirectory 
+            && !this.isRoot
+        ) {
```
Vegita2: if condition should be refactored into it's own method. It is the same as above, just negated

2767mr:
Btw, this also fixes #111

webapp/src/app/components/load-map/virtualMapNode.model.ts
2767mr: Fixed in 9a3a38d

PR #150, opened 2020-03-23
Added entity filter

### LINKED ISSUE #35: Show/hide entities based upon specific conditions
It's hard to edit the map when everything is thrown in your face all at once. Being able to filter based upon a specific condition would be nice.

2767mr:
Closes #35


webapp/src/app/shared/phaser/entities/cc-entity.ts
```diff
@@ -569,4 +571,29 @@ export abstract class CCEntity extends BaseObject {
...
 		
 	}
+
+	protected filter(filter: string): boolean {
+		const lower = filter.toLocaleLowerCase();
+		const attributes = this.getAttributes();
+
+		for (const name of Object.keys(attributes)) {
+			const toLowerCase = (this.details.settings[name] || '').toLowerCase;
+			if (toLowerCase && (this.details.settings[name] || '').toLowerCase().includes(lower)) {
```
Vegita2: Is this a string check? Why not use `typeof x === 'string'` ?

webapp/src/app/shared/phaser/entities/cc-entity.ts
```diff
@@ -94,6 +94,8 @@ export abstract class CCEntity extends BaseObject {
 		this.details = <any>{
 			type: typeName
 		};
+		
+		Globals.globalEventsService.filterEntity.subscribe(filter => this.setVisible(this.filter(filter)));
```
Vegita2: subscription should be unsubscribed when entity is destroyed

2767mr:
> currently selected item keeps selection even if it is filtered, not sure if this should be changed or not
I think that should be kept as is because it gives the user more options. At the moment it the filter is hidden anyway so you can't even change it.

> I would prefer if the filter input is not visible if an entity is selected
I added that but I think it's really counterintuitive. I also tried making it smaller but I failed. I would appreciate if you'd change the design like you want it to be. As long as there is a textbox where you can enter the filter I am fine with it.

Vegita2:
I played around with it a bit and I still prefer the field to be hidden if an entity is selected. We could add this as an option in the settings page so everyone can decide for themselves. 

Another alternative i thought of is to completely move the filter to the toolbar, because it is not limited to the entities tab

PR #151, opened 2020-03-24
Added tileset select

### LINKED ISSUE #68: Select tileset from dropdown instead of text input
Currently, the user has to enter the full path of an tileset in order to use it.
The editor should:
* Provide a selection of existing tilesets.
* Suggest appropriate tilemaps for known types of layers such as 'collisiontiles-16x16
' for collision layers.

### LINKED ISSUE #69: indicate missing/broken tileset
The game crashes when trying to load maps that have layers without/broken tilesets. Map editor should show a warning/error in that case

2767mr:
Closes #68


2767mr:
The idea was that there is no way to enter an invalid tileset. I guss this is not the case here, since an empty tileset is invalid. I'll remove it from the description.

PR #152, opened 2020-03-24
Fixed prop not showing bugs

2767mr:
If a `prop` has `SUB` on `Anims` it seems like all of the `SUB` entries should be displayed. However, this seems to cause some bugs because `renderType: 'lighter'` doesn't seem to work properly.
For now, this PR changes props to draw the first entry which always seems to be the base of the entity.

Example:
image

### DIFF
--- a/webapp/src/app/shared/phaser/entities/registry/prop.ts
+++ b/webapp/src/app/shared/phaser/entities/registry/prop.ts
@@ -222,7 +222,7 @@ export class Prop extends CCEntity {
 		
 		if (anims.sheet || anims.namedSheets) {
 			this.setupAnim(settings, anims, sheetDef, propDef.name, sprites);
-		} else { //<- temporary
+		} else {
 			if (anims.SUB) {
 				for (const anim of anims.SUB as Anims[]) {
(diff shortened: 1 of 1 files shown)


webapp/src/app/shared/phaser/entities/registry/prop.ts
```diff
@@ -218,7 +218,67 @@ export class Prop extends CCEntity {
 			offset?: Point3;
 			renderMode?: string
 		}[] = [];
+
+
+		if (anims.sheet || anims.namedSheets) {
+			this.setupAnim(settings, anims, sheetDef, propDef.name, sprites);
+		} else { //<- temporary
```
Vegita2: why temporary?

2767mr: Because some entities have a `sheet` as well as `SUB` entries and I think they are supposed to be rendered as if `sheet` was inserted as the first element into `SUB`.
Since `renderMode` has been fixed the else should be removed.

2767mr: Turns out that fixing that is a bit more complex and involves handeling `frame` entries. The current implementation should give a result that is good enough for an editor.

PR #156, opened 2020-04-17
Changed tileset from select to autocomplete

2767mr:
Due to some requests from users I changed the tileset select to an autocomplete.


Vegita2:
Where was that requested?

Anyway i tried the change and It doesn't work on my end. No matter what I type in that field the autocomplete list is not filtered.

2767mr:
https://discordapp.com/channels/382339402338402315/383264360518123520/700604011681677363

The second one, which was by @ac2pic (Hoyori#8777) got deleted.

ac2pic:
I can confirm that I suggested the change.

PR #158, opened 2020-04-20
Mod support

### LINKED ISSUE #54: Add mod support
Pretty much simulate the patch system.

2767mr:
* Does not work for ccmods
* Fixes #54 
* Closes #156 because custom can only be located inside vanilla or mod folders

### DIFF
--- a/backend/src/config.ts
+++ b/backend/src/config.ts
@@ -1,4 +1,5 @@
 export let config = {
-	pathToCrosscode: 'C:/Program Files (x86)/Steam/steamapps/common/CrossCode/assets/'
+	pathToCrosscode: 'C:/Program Files (x86)/Steam/steamapps/common/CrossCode/assets/',
+	mod: ''
 };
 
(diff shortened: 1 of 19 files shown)


Vegita2:
really cool addition, i played around with it a bit and everything seems to work.

Would it be possible to make the mod change not require a restart? 
I would also like to have the mod selection in the settings page when using the web version. The config.ts file is not a good solution. It works for the path because it doesn't require changes often and we had to implement a custom dialog to select the directory. 

The settings page is fine for the mod selection, but I would increase the spacing from the top a bit and make the select field smaller (setting a reasonable max width, so smaller windows can still see the input). Maybe also add a small description similar to the assets folder field

2767mr:
> Would it be possible to make the mod change not require a restart?

As far as I know, there is no easy way to replace textures without reloading the entire map editor. Of course, we could do this automatically instead of having the user restart it manually.

> I would also like to have the mod selection in the settings page when using the web version. The config.ts file is not a good solution. It works for the path because it doesn't require changes often and we had to implement a custom dialog to select the directory.

This is a leftover from when I started working on mod support because then you had to select the directory instead of the mod name. I'll look into changing it.

PR #167, opened 2020-06-08
Update setup

2767mr:
* Updated node dependencies
* Changed `npm install` to `npm ci` in travis config for more reliable builds
* Changed all typescript targets to `ES2018` for consistency
* Changed `tslint` to `eslint` because `tslint` is deprecated
  * This comes with rule changes that need to be reviewed


Vegita2 (changes requested):
I don't like the no unused vars rule. The function gets called with the parameter anyway so removing it just makes it difficult to edit it in the future. `ngOnChanges` should also always have the `SimpleChanges` parameter.

For the other rules I'd suggest just changing them as we encounter them.

2767mr:
> For the other rules I'd suggest just changing them as we encounter them.

Does that mean it is fine as is?

Vegita2:
I looked through all the changes and they look good. So if it compiles we can merge it. There are probably more things that break with the updates but the basic things ahould work

Vegita2 (approved):

PR #169, opened 2020-06-13
Npc and freeze fix

### LINKED ISSUE #168: NPC entities not displaying
(Tested on 3 clients)
Currently attempting to create your own NPC entity will result with no reaction. Copying a existing NPC entity will work fine, until the NPC is edited, from then on the editor will act as though it is not there when re loading the map.

2767mr:
Fixes #168 
* Changed `Helper.getJson` to return undefined if no file was found.
* Fixed infinite loop in SideNav.

### DIFF
--- a/webapp/src/app/components/sidenav/sidenav.component.html
+++ b/webapp/src/app/components/sidenav/sidenav.component.html
@@ -1,5 +1,5 @@
 <div fxLayout="column" class="nav-container" *ngIf="tilemap">
-	<mat-tab-group (selectedTabChange)="tabChanged($event)" [(selectedIndex)]="activeTab">
+	<mat-tab-group (selectedTabChange)="tabChanged($event)" [selectedIndex]="activeTab">
 		<mat-tab label="Layers">
 			<app-layers class="hacky-tab-height-fix" fxLayout="column"></app-layers>
 		</mat-tab>
--- a/webapp/src/app/components/sidenav/sidenav.component.html
+++ b/webapp/src/app/components/sidenav/sidenav.component.html
@@ -1,5 +1,5 @@
 <div fxLayout="column" class="nav-container" *ngIf="tilemap">
-	<mat-tab-group (selectedTabChange)="tabChanged($event)" [selectedIndex]="activeTab">
+	<mat-tab-group (selectedTabChange)="tabChanged($event)" [(selectedIndex)]="activeTab">
 		<mat-tab label="Layers">
 			<app-layers class="hacky-tab-height-fix" fxLayout="column"></app-layers>
 		</mat-tab>
--- a/webapp/src/app/components/sidenav/sidenav.component.ts
+++ b/webapp/src/app/components/sidenav/sidenav.component.ts
@@ -35,22 +35,11 @@ export class SidenavComponent implements OnInit {
 		this.mapLoader.tileMap.subscribe(tilemap => {
 			this.tilemap = tilemap;
 			const currentView = this.globalEvents.currentView;
-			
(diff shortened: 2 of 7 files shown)


webapp/src/app/components/sidenav/sidenav.component.ts
```diff
@@ -35,22 +35,11 @@ export class SidenavComponent implements OnInit {
...
-			this.currentView = view;
-			switch (view) {
-				case EditorView.Layers:
-					this.activeTab = 0;
-					break;
-				case EditorView.Entities:
-					this.activeTab = 1;
-					break;
+			if (view !== undefined && this.activeTab != view) {
+				this.activeTab = view;
```
Vegita2, line 42: Missed this the first time. `activeTab` is a number and view is not, should be changed to match.

2767mr: Fixed

webapp/src/app/components/sidenav/sidenav.component.ts
```diff
@@ -14,7 +14,7 @@ import {MatTabChangeEvent} from '@angular/material';
 })
 export class SidenavComponent implements OnInit {
 	
-	activeTab = 0;
+	activeTab = EditorView.Layers;
```
Vegita2, line 17: If `activeTab` is changed to EditorView then it is the same as `currentView`. In this case the `currentView` variable could be removed

webapp/src/app/components/sidenav/sidenav.component.html
```diff
@@ -1,5 +1,5 @@
 <div fxLayout="column" class="nav-container" *ngIf="tilemap">
-	<mat-tab-group (selectedTabChange)="tabChanged($event)" [selectedIndex]="activeTab">
+	<mat-tab-group (selectedTabChange)="tabChanged($event)" [(selectedIndex)]="activeTab">
```
Vegita2, line 2: Is the `selectedIndex` event really necessary? In this case activeTab is set twice, once from `(selectedTabChange)` and once from `(selectedIndex)`

webapp/src/app/components/sidenav/sidenav.component.ts
```diff
@@ -35,22 +35,11 @@ export class SidenavComponent implements OnInit {
...
 		this.globalEvents.currentView.subscribe(view => {
-			this.currentView = view;
-			switch (view) {
-				case EditorView.Layers:
-					this.activeTab = 0;
-					break;
-				case EditorView.Entities:
-					this.activeTab = 1;
-					break;
+			if (view !== undefined && this.activeTab != view) {
```
Vegita2, line 41: lint: ` != should be !==`

webapp/src/app/components/sidenav/sidenav.component.ts
2767mr: `currentView` is in fact unused. Fixed by removal in fdd25bafcbf6f475002ed4916aa8757f3d4bd3cc.

webapp/src/app/components/sidenav/sidenav.component.ts
2767mr: Fixed in 091d376bb3c9cdc9d1c93f4672df0b4a50ef2107

webapp/src/app/components/sidenav/sidenav.component.html
2767mr: The binding seems to be unnecessary. Removed in 6ead0628ebbbc7ece8c82c7d0f47a99749600658.

PR #178, opened 2020-09-18
generate unique mapId for new entities

### LINKED ISSUE #176: Generate mapIds for EventTrigger
The game uses these in order to determine whether or not it has been triggered.

Vegita2:
fixes #176

### DIFF
--- a/webapp/src/app/shared/phaser/entities/cc-entity.ts
+++ b/webapp/src/app/shared/phaser/entities/cc-entity.ts
@@ -1,12 +1,12 @@
-import {CCMap} from '../tilemap/cc-map';
-import {MapEntity, Point, Point3} from '../../../models/cross-code-map';
-import {Helper} from '../helper';
+import { CCMap } from '../tilemap/cc-map';
+import { MapEntity, Point, Point3 } from '../../../models/cross-code-map';
+import { Helper } from '../helper';
 import * as Phaser from 'phaser';
-import {Vec2} from '../vec2';
+import { Vec2 } from '../vec2';
 
-import {Globals} from '../../globals';
-import {BaseObject} from '../base-object';
-import {Subscription} from 'rxjs';
+import { Globals } from '../../globals';
+import { BaseObject } from '../base-object';
+import { Subscription } from 'rxjs';
 
 export interface ScaleSettings {
 	scalableX: boolean;
@@ -28,6 +28,12 @@ export interface AttributeValue {
 	optional?: boolean;
 	
 	// TODO: not needed anymore, cc source is not obfuscated
+	[key: string]: any;
+}
+
+export interface DetailSettings {
+	mapId: number;
+	
 	[key: string]: any;
 }
 
@@ -72,13 +78,19 @@ export abstract class CCEntity extends BaseObject {
 	get selected(): boolean {
 		return this._selected;
 	}
-
+	
 	// drag
 	public isDragged = false;
 	public startOffset: Point = {x: 0, y: 0};
 	
 	// zIndex: number;
-	details: { level: { level: number, offset: number }, type: string, settings: any } = <any>{};
+	details: {
+		level: {
+			level: number;
+			offset: number;
+		}; type: string;
+		settings: DetailSettings;
+	} = <any>{};
 	entitySettings: {
 		collType: string;
 		baseSize: Point3;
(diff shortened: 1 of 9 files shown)


webapp/src/app/shared/phaser/entities/cc-entity.ts
```diff
@@ -31,6 +31,12 @@ export interface AttributeValue {
 	[key: string]: any;
 }
 
+export interface DetailSettings {
+	mapId: number;
```
2767mr, line 35: All uses of `mapId` act as if it were optional. One even does `Partial<DetailSettings>` just for that.

webapp/src/app/shared/phaser/entities/cc-entity.ts
```diff
@@ -95,7 +107,7 @@ export abstract class CCEntity extends BaseObject {
 		pivot: Point;
 	} = <any>{};
 	
-	protected constructor(scene: Phaser.Scene, map: CCMap, x: number, y: number, typeName: string) {
+	public constructor(scene: Phaser.Scene, map: CCMap, x: number, y: number, typeName: string) {
```
2767mr, line 110: Why was this changed? Seeing the commet of `getEntity` this might be a leftover?

webapp/src/app/shared/phaser/entities/entity-manager.ts
```diff
@@ -296,12 +283,36 @@ export class EntityManager extends BaseObject {
...
+		}, true);
+		
+		return e;
+	}
+	
 	async generateEntity(entity: MapEntity): Promise<CCEntity> {
 		const entityClass = Globals.entityRegistry.getEntity(entity.type);
-		
-		const ccEntity = new entityClass(this.scene, this.map, entity.x, entity.y, entity.type);
+		console.assert(this.map);
```
2767mr, line 308: Add description to assert

webapp/src/app/shared/phaser/entities/registry/entity-registry.service.ts
```diff
@@ -1,45 +1,43 @@
...
+	private defaultEntity = DefaultEntity;
 	
 	constructor() {
-		this.setDefaultEntity(DefaultEntity);
 		this.register('Prop', Prop);
 		this.register('ScalableProp', ScalableProp);
 		this.register('ItemDestruct', ItemDestruct);
 		this.register('Destructible', Destructible);
 		this.register('NPC', NPC);
+		this.register('NPC', NPC);
```
2767mr, line 22: This looks like it was added by accident

webapp/src/app/shared/phaser/entities/registry/entity-registry.service.ts
```diff
@@ -1,45 +1,43 @@
...
 	public getAll() {
 		return this.entities;
 	}
 	
-	public getEntity(type: string): new (...args: any[]) => CCEntity {
-		return this.entities[type] || this.defaultEntity;
+	// typed as DefaultEntity so constructor can be used with parameter checking.
+	// CCEntity is abstract and doesn't allow using the constructor
+	public getEntity(type: string): typeof DefaultEntity {
+		return (this.entities[type] || this.defaultEntity) as any;
```
2767mr, line 41: Why not use `DefaultEntity` directly? 
```suggestion
		return (this.entities[type] ?? DefaultEntity) as typeof DefaultEntity
```
(I don't care of one does `??` or `||` in this case, just a suggestion)

webapp/src/app/shared/phaser/tilemap/cc-map.ts
```diff
@@ -88,8 +89,16 @@ export class CCMap {
 			this.inputLayers = undefined;
 		}
 		
+		this.lastMapId = 0;
```
2767mr, line 92: This is set to `0` while it is set to `1` at the declaration above

2767mr (changes requested):
If entities are copied, the `mapId` is also copied.

webapp/src/app/shared/phaser/entities/cc-entity.ts
Vegita2: I moved the call to `map.getUniqueMapid()` to better reflect what i meant with that. It basically says imported map can contain entities without a mapId, but the internal representation of entities always has a mapId.

It is currently unused but having the type is still better than using `any`

webapp/src/app/tests/test-helper.ts
```diff
@@ -0,0 +1,33 @@
...
+		return new Promise(((res) => {
+			let imported: CrossCodeMap;
+			let sub: Subscription;
+			// eslint-disable-next-line prefer-const
+			sub = service.tileMap.subscribe(map => {
+				if (!map) {
+					return;
+				}
+				
+				sub!.unsubscribe();
```
2767mr, line 19: There seems to be a some sort of race condition or test ordering issue here because one test failed because `sub` was undefined. Since I couldn't reprocude the issue I suggest making a note/issue and investigating later.

2767mr (approved):

PR #179, opened 2020-09-18
Layer bugfix

### LINKED ISSUE #165: layer generation adds wrong tiles
Discord: WatDuhHekBro#7158

Well I found out the problem. For some strange reason, when I created my collision maps, the map editor generated them with empty values of -1 instead of 0, which is what caused all my collision problems.

image

Vegita2:
- Initialize tiles with 0 instead of -1 in new layers, fixes #165
- Resize and offset should now work properly again
- Fixed: changing layer tileset causes previously made changes to be lost


2767mr (changes requested):
Should we replace all `.getPhaserLayer()!` with `.getPhaserLayer()`?

2767mr (approved):

PR #189, opened 2020-10-10
improved tile drawing

### LINKED ISSUE #183: Drawing tiles fast leaves some holes
When moving the mouse fast during tile drawing some tiles on the way don't get updated

Vegita2:
Improved tile drawing by drawing a line of points instead of a single point. 
Avoids holes when drawing fast.
lines

Fixes #183


2767mr (approved):
A very nice QoL addition!

There is no changelog entry yet.

Issue #173, opened 2020-07-14
Refactor event editor


Vegita2:
Requested by Bakafish#0001

- Being able to add/delete steps in action/event step lists more intuitively. I actually have no idea how to delete said steps at the moment.
- Mod support with custom action/event steps would be handy, although probably not very easy to implement.

Vegita2:
additional nice to have:
- use arrow keys to move up/down in the events list
- drag and drop events
- after deleting an event, automatically select the next one
- after adding a new event, select the event and open it's detail page
- allow selecting multiple events with shift/ctrl to mass copy/delete
- allow double click to add a new event on all of the empty space (blue border) instead of only the line (green border)

![image](https://user-images.githubusercontent.com/9483499/87465926-b61a2100-c615-11ea-8da4-682a1a3db28a.png)

Vegita2:
good addition. I wasn't really sure at first about having the details view always visible, but after working on my mod a bit this really makes it faster/easier to get events done.
But I think the ui/ux needs some work before we should merge it and I also noticed some smaller bugs.

- opening DO_ACTION action list opens another event editor with the same details panel, so it can open itself infinite times through the already opened details panel
- default detail panel size should be smaller, just enough to fit the input fields. the list should have the additional space
- event text doesn't overflow. This looks cleaner but at the same time can hide important information. Maybe add a toggle?
- right click shouldn't open default browser context menu
- explicit delete button for events (not sure if this is necessary)
- revert button for events
- changing IF to include/exclude an ELSE block doesn't show it in the events list, only if something else is changed
- selecting the last item / an empty item doesn't deselect the previous selection
- selecting a row doesn't select the other options associated with it (if, show_choice). Would make it easier to see what exactly gets deleted/moved/copied 
- overlay header/toolbar casts a shadow over event detail, but not on the event list (event widget)

drag drop issues
- last item / empty item is still moved around on drag even if it makes no difference. It's difficult to see where the item is going to be in if/else blocks 
- dropped item moves to wrong position visually
- it looks a lot better if the event is first selected before moving it around, maybe just autoselect it on move
- when dragging long rows they show multiple lines, intended? If yes it should be slightly transparent because it hides too much of the list  
- dragging IF should drag the whole block

- X and Event title should be on the same line, I think it looks better
![same_line_x_title](https://user-images.githubusercontent.com/9483499/99306080-ee946a80-2854-11eb-96e7-c2788de3a283.png)

[...]

PR #190, opened 2020-11-13
Reworked event editor

2767mr:
The event editor now shows events and event details at the same time.
It is possible to edit details and see changes in real time.
Events can now be reordered through drag and drop.


Vegita2 (changes requested):
good addition. I wasn't really sure at first about having the details view always visible, but after working on my mod a bit this really makes it faster/easier to get events done.
But I think the ui/ux needs some work before we should merge it and I also noticed some smaller bugs.

- opening DO_ACTION action list opens another event editor with the same details panel, so it can open itself infinite times through the already opened details panel
- default detail panel size should be smaller, just enough to fit the input fields. the list should have the additional space
- event text doesn't overflow. This looks cleaner but at the same time can hide important information. Maybe add a toggle?
- right click shouldn't open default browser context menu
- explicit delete button for events (not sure if this is necessary)
- revert button for events
- changing IF to include/exclude an ELSE block doesn't show it in the events list, only if something else is changed
- selecting the last item / an empty item doesn't deselect the previous selection
- selecting a row doesn't select the other options associated with it (if, show_choice). Would make it easier to see what exactly gets deleted/moved/copied 
- overlay header/toolbar casts a shadow over event detail, but not on the event list (event widget)

drag drop issues
- last item / empty item is still moved around on drag even if it makes no difference. It's difficult to see where the item is going to be in if/else blocks 
- dropped item moves to wrong position visually
- it looks a lot better if the event is first selected before moving it around, maybe just autoselect it on move
- when dragging long rows they show multiple lines, intended? If yes it should be slightly transparent because it hides too much of the list  
- dragging IF should drag the whole block

- X and Event title should be on the same line, I think it looks better
same_line_x_title

- event editor is a bit small when editing NPC's and the background of the details panel is too similar to the rest of the overlay. I'd suggest something like that
image

Also here are the feature requests from the issue, can you comment on them on how difficult you think it is to implement them and if it makes sense to add it here, another time or not at all.

- [ ] Being able to add/delete steps in action/event step lists more intuitively. I actually have no idea how to delete said steps at the moment.
- [ ] Mod support with custom action/event steps would be handy, although probably not very easy to implement.
- [ ] use arrow keys to move up/down in the events list
- [X] drag and drop events
- [ ] after deleting an event, automatically select the next one
- [X] after adding a new event, select the event and open it's detail page
- [ ] allow selecting multiple events with shift/ctrl to mass copy/delete
- [ ] allow double click to add a new event on all of the empty space (blue border) instead of only the line (green border)

image

Edit: additional nice to haves
- transition sidepanel in/out
- resizeable details panel
- detachable details panel

2767mr:
> default detail panel size should be smaller

I made it big because JSON needs a lot of space. Should we just ignore that? Also, I thought it was more important to display long texts in the details page rather than the list.

> event text doesn't overflow

It doesn't because then the strips in the background would look really weird as two different colors would still be the same element.

> right click shouldn't open default browser context menu

This is an issue I noticed outside of the event editor too. I'll look into it.

> explicit delete button for events

I intend to add a couple of new buttons after this PR in order to further improve it as described in #173. I agree that a delete button would be safe to add in this PR (maybe in the top right of the details?) but other buttons such as add, move, cut, copy and paste are more difficult to place.

> revert button for events

Do you mean something like CTRL + z for events? If so, I'd like to add it in a later PR.

> selecting the last item / an empty item doesn't deselect the previous selection

Even though it may seem weird at first I think this should be left as is. There is no real reason why one would deselect everything by clicking on an empty entry and if they did so by mistake it is probably best to act as if nothing happened. The only thing I was thinking about changing whether clicking on "else" should select the "if".

> dragging IF should drag the whole block

This is an issue with the dragging framework because it doesn't really support trees. The illusion of a tree is actually manually made using spaces at the beginning of each element. I'll try to fix this by crafting the dragged element by hand to include the children as well as manally add invisibility to elements not actually dragged.

> Being able to add/delete steps in action/event step lists more intuitively. I actually have no idea how to delete said steps at the moment.

As mentioned above, I'd like to add some buttons in a future PR as this is a GUI issue that probably needs to be discussed. I'll add a delete button in this PR.

> Mod support with custom action/event steps would be handy, although probably not very easy to implement.

Mod support will require some more ground work I don't want to include in this PR as it will probably touch a few other systems.

> use arrow keys to move up/down in the events list
> after deleting an event, automatically select the next one

I'll try to add it to this PR. I need to look into some selection related bugs anyway.

> allow selecting multiple events

Although I think it'll feel a bit awkward to use I'll try to implement it. It will probably close the detail page when doing so since I have no idea what to do instead.

> allow double click to add a new event on all of the empty space 

I'll add that.

Vegita2:
> I made it big because JSON needs a lot of space. Should we just ignore that? Also, I thought it was more important to display long texts in the details page rather than the list.

JSON inputs should always be ignored. They are only a temporary solution and editing them in the editor is quite bad. Currently it is much better to copy paste them instead of editing it directly in the editor.

> It doesn't because then the strips in the background would look really weird as two different colors would still be the same element.

The dot on the left indicates a new event, the background just differentiates between the rows. We had it like that before and I had no problems with it. I also don't have a strong opinion regarding this so both ways are ok with me.   

> Do you mean something like CTRL + z for events? If so, I'd like to add it in a later PR.

Yes. imagine you edit an event that contains some json and it takes a while, then you realize that this is the wrong event step and you want to revert it, but keep the changes in the other event steps. Similar to the old cancel button

> Even though it may seem weird at first I think this should be left as is. There is no real reason why one would deselect everything by clicking on an empty entry and if they did so by mistake it is probably best to act as if nothing happened. The only thing I was thinking about changing whether clicking on "else" should select the "if".

If it happens by mistake, the user can simply click on the event again. I also don't see how this could happen by mistake, if you click on other events it also deselects the current one so there is not much of a difference. Also the problem with this current behavior is that it's not clear what you currently have selected. If you want to add a new event step there is no indication where it's going to be inserted.

> Mod support will require some more ground work I don't want to include in this PR as it will probably touch a few other systems.

I think mod support should be completely independent from the event editor. The editor just handles events given to it, but what event steps actually exist is defined somewhere else.

Vegita2:
found new stuff

- It's not possible to copy paste the same event multiple times, after the first paste it just doesn't work anymore
- if opening the event details and closing the npc/event editor with the cancel button, the detail view keeps it's state and shows the same event again no matter which event editor is opened. step-by-step: select entity and click on event editor widget -> click on an event step to open the detail view -> click cancel button -> select another entity -> open event editor widget -> editor now shows the previously selected event even if it was in another entity  

Should be the same as 
> opening DO_ACTION action list opens another event editor with the same details panel, so it can open itself infinite times through the already opened details panel

If the detail page resets state after closing both of those should be fixed.

Vegita2 (changes requested):
bugs:
- no change detection triggered after adding new event, move loadEvent function to setter and make component onpush

- after dragging event, changes in the detail page will not reflect in the event list

- I don't understand how ctrl + z behaves. If I change 2 events, ctrl+z only reverts the changes on the second element, but not the first no matter how often i press. But if I select another element it now somehow works. Also adding a new element or deleting one messes everything up. I'd suggest also reverting the selected element with ctrl + z so it is easier to understand what exactly happens.

- changing IF to include/exclude an ELSE block doesn't show it in the events list, only if something else is also changed

nice to haves (I'll add a new issue for that)
- remember width of event details page on next start. Currently after closing/opening the event editor it's back to default (50%) #194

Vegita2 (approved):
code looks good, just one issue left that I found

- when navigating with arrow keys, the last event is still selected

PR #198, opened 2020-12-28
automate release process

### LINKED ISSUE #195: automate release process
while we are moving to github actions (#193) we could automate the whole release process.

- We need to automatically update the version number in the package.json/package-lock.json
- update the changelog titles `[Unreleased]` should be changed to the current version
- make a new release with the changelog text added to it
- build and add binaries to the release

might be useful to implement this
https://github.com/dangoslen/changelog-enforcer/
https://github.com/mindsers/changelog-reader-action
https://github.com/marketplace/actions/automated-version-bump

Vegita2:
- builds the project on every push except on master branch
- pull request also run a check to see if the changelog was changed

prepare release:
- is triggered when a pull request is merged and has a label `bump:x`
- changelog is automatically updated
- package.json and package-lock.json version is updated
- a tag for the version is created
- pushes changes to the repo

release:
- is triggered when a new tag is pushed. Can be manually or through the prepare release step
- builds the binaries on linux, mac and windows
- makes a new release with the binaries and appends the changelog to it
   
   
------
  
This pr requires:  
- 3 new labels `bump:major` `bump:minor` `bump:patch`
- a github personal access token with push access to this repo stored as a secret with the name `PAT`

closes #195

### DIFF
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -1,0 +1,55 @@
+name: prepare_release
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  check:
+    name: check release version
+    runs-on: ubuntu-latest
+    outputs:
+      skip: ${{ steps.bumpr.outputs.skip }}
+      version: ${{ steps.bumpr.outputs.next_version }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: bumpr
+        uses: haya14busa/action-bumpr@v1
+        with:
+          dry_run: true
+  prepare:
+    name: prepare release
+    needs: check
+    if: '!needs.check.outputs.skip'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+      - name: setup node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - name: update changelog
+        run: |
+          export NEXT_VERSION_V=${{ needs.check.outputs.version }}
+          export NEXT_VERSION=${NEXT_VERSION_V:1}
+          sed -i "/\[Unreleased\]/a ## [$NEXT_VERSION] newDateHere" CHANGELOG.md
+          sed -i "s/newDateHere/$(date '+%Y-%m-%d')/g" CHANGELOG.md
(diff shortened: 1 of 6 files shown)


.github/workflows/prepare_release.yml
```diff
@@ -0,0 +1,55 @@
...
+          fetch-depth: 0
+      - name: setup node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - name: update changelog
+        run: |
+          export NEXT_VERSION_V=${{ needs.check.outputs.version }}
+          export NEXT_VERSION=${NEXT_VERSION_V:1}
+          sed -i "/\[Unreleased\]/a ## [$NEXT_VERSION] newDateHere" CHANGELOG.md
```
2767mr, line 38: Wouldn't it make sense to use another env variable instead of `newDateHere`?

.github/workflows/prepare_release.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: make version tag
+        working-directory: webapp
+        run: |
+          mkdir .git
+          npm version ${{ needs.check.outputs.version }}
+      - name: Push changes
+        uses: ad-m/github-push-action@master
+        with:
+          tags: true
+          github_token: ${{ secrets.PAT}}
```
2767mr, line 54: This secret needs to be created (or changed to `GITHUB_TOKEN`).
Also, missing whitespace.

2767mr (changes requested):
I tried this on a fork and it didn't work even with changelog and token - the `release` action didnt trigger. See: https://github.com/2767mr/crosscode-map-editor

.github/workflows/prepare_release.yml
Vegita2: Maybe. But I found the code to replace a string with a date on stackoverflow and used that. I'm not really good with command line stuff in general

.github/workflows/prepare_release.yml
Vegita2: Github token doesn't work because it doesn't trigger another workflow. It requires a personal access token

2767mr: I'll aprove the PR once the secret has been created

2767mr (approved):

PR #207, opened 2021-02-07
Add map pan key control

bawood2:
Adds `SHIFT` as a control to pan the camera,  allowing a user working on a touch pad to perform the already existing zoom (ALT + two finger zoom) and what this pull request covers, to pan ( SHIFT + mouse click finger move)

### DIFF
--- a/webapp/src/app/shared/phaser/entities/entity-manager.ts
+++ b/webapp/src/app/shared/phaser/entities/entity-manager.ts
@@ -149,16 +149,16 @@ export class EntityManager extends BaseObject {
 					this.selectedEntities.forEach(entity => {
 						entity.isDragged = false;
 					});
+
+					// if panning do not deselect entities
+					if (Globals.panning) {
+						return;
+					}
 					
 					if (this.gameObjectDown) {
 						this.gameObjectDown = false;
 					} else {
 						const entities = this.selectionBox.onInputUp();
-
-						// if panning do not deselect entities
-						if (Globals.panning) {
-							return;
-						}
 
 						if (!this.multiSelectKey.isDown) {
 							this.selectEntity();
--- a/webapp/src/app/shared/phaser/entities/entity-manager.ts
+++ b/webapp/src/app/shared/phaser/entities/entity-manager.ts
@@ -112,6 +112,11 @@ export class EntityManager extends BaseObject {
 					entity = gameObject[0].getData('entity') as CCEntity;
 				}
 				
+				// if panning return once entity as been selected to prevent dragging of selected entities
+				if (Globals.panning) {
+					return;
+				}
+
 				if (entity) {
 					this.gameObjectDown = true;
 					
@@ -140,6 +145,12 @@ export class EntityManager extends BaseObject {
 					this.selectEntity();
 					this.showAddEntityMenu();
 				} else if (pointer.leftButtonReleased()) {
+
+					// if panning do not deselect entities
+					if (Globals.panning) {
+						return;
+					}
+
 					this.selectedEntities.forEach(entity => {
 						entity.isDragged = false;
 					});
--- a/webapp/src/app/shared/phaser/entities/entity-manager.ts
+++ b/webapp/src/app/shared/phaser/entities/entity-manager.ts
@@ -144,12 +144,7 @@ export class EntityManager extends BaseObject {
 				if (pointer.rightButtonReleased()) {
 					this.selectEntity();
 					this.showAddEntityMenu();
-				} else if (pointer.leftButtonReleased()) {
-
-					// if panning do not deselect entities
-					if (Globals.panning) {
-						return;
-					}
+				} else if (pointer.leftButtonReleased()) {			
 
 					this.selectedEntities.forEach(entity => {
 						entity.isDragged = false;
@@ -159,6 +154,12 @@ export class EntityManager extends BaseObject {
 						this.gameObjectDown = false;
 					} else {
 						const entities = this.selectionBox.onInputUp();
+
+						// if panning do not deselect entities
+						if (Globals.panning) {
+							return;
+						}
+
 						if (!this.multiSelectKey.isDown) {
 							this.selectEntity();
 						}
--- a/webapp/src/app/shared/phaser/map-pan.ts
+++ b/webapp/src/app/shared/phaser/map-pan.ts
@@ -81,11 +81,11 @@ export class MapPan extends Phaser.GameObjects.GameObject {
 	
 	preUpdate() {
 
-		Globals.panning = this.panKey.isDown || this.scene.input.activePointer.middleButtonDown();
-
 		const pointer = this.scene.input.activePointer;
+		
+		Globals.panning = this.panKey.isDown || pointer.middleButtonDown();
 
-		if (Globals.panning && this.isScrolling && this.active) {
+		if (Globals.panning && this.isScrolling && this.active && pointer.isDown) {
 			const dx = pointer.x - this.startMouse.x;
 			const dy = pointer.y - this.startMouse.y;
 			
--- a/webapp/src/app/shared/phaser/map-pan.ts
+++ b/webapp/src/app/shared/phaser/map-pan.ts
@@ -81,11 +81,7 @@ export class MapPan extends Phaser.GameObjects.GameObject {
 	
 	preUpdate() {
 
-		if (this.panKey.isDown || this.scene.input.activePointer.middleButtonDown()) {
-			Globals.panning = true;
-		} else {
-			Globals.panning = false;
-		}
+		Globals.panning = this.panKey.isDown || this.scene.input.activePointer.middleButtonDown();
 
 		const pointer = this.scene.input.activePointer;
 
--- a/webapp/src/app/shared/phaser/map-pan.ts
+++ b/webapp/src/app/shared/phaser/map-pan.ts
@@ -8,11 +8,13 @@ export class MapPan extends Phaser.GameObjects.GameObject {
 	private startCam: Point = {x: 0, y: 0};
 	
 	private zoomKey: Phaser.Input.Keyboard.Key;
+	private panKey: Phaser.Input.Keyboard.Key;
 	
 	constructor(scene: Phaser.Scene, type: string) {
 		super(scene, type);
 		
 		this.zoomKey = scene.input.keyboard.addKey(Phaser.Input.Keyboard.KeyCodes.ALT, false);
+		this.panKey = scene.input.keyboard.addKey(Phaser.Input.Keyboard.KeyCodes.CTRL, false);
 		// game.input.mouseWheel.callback = (v) => this.onMouseWheel(v);
 		scene.input.on('wheel', (
 			pointer: Phaser.Input.Pointer,
@@ -29,6 +31,15 @@ export class MapPan extends Phaser.GameObjects.GameObject {
 		if (!this.active) {
 			return;
 		}
+
+		// set global panning state when the panKey is down
+		if (this.panKey.isDown || this.scene.input.activePointer.middleButtonDown()) {
+			Globals.panning = true;
+		} else {
+			Globals.panning = false;
+			return;
+		}
+
 		this.isScrolling = true;
 		const cam = this.scene.cameras.main;
 		Vec2.assign(this.startMouse, this.scene.input.activePointer);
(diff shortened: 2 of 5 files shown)


bawood2:
My only question is https://github.com/CCDirectLink/crosscode-map-editor/pull/207/commits/cad758938efb947698663882a5005e3961d63ec5#diff-32eb581b52594520034d052c9e7ce1c7950ba88d11baa800be941356b587c7aeR186

What is this `shiftKey.isDown` supposed to be doing?  In my initial commit, I ran into the issue where when I was panning, tiles were being drawn.  I added the pan key tile drawer and a check here https://github.com/CCDirectLink/crosscode-map-editor/pull/207/commits/cad758938efb947698663882a5005e3961d63ec5#diff-32eb581b52594520034d052c9e7ce1c7950ba88d11baa800be941356b587c7aeR163, but is that going to conflict at all with the pre existing `shiftKey.isDown`?

webapp/src/app/shared/phaser/tilemap/tile-drawer.ts
```diff
@@ -156,6 +158,12 @@ export class TileDrawer extends BaseObject {
 			if (this.lastDraw.x === p.x && this.lastDraw.y === p.y) {
 				return;
 			}
+
+			// skip drawing if panning key is down
+			if (this.panKey.isDown) {
+				return;
+			}
+
```
2767mr, line 166: This clashes with the existing shift key functionality. Before this PR the shift key would allow one to draw without the auto tile feature interfering. I could see the old shift key functionality being rebound to CTRL.

bawood2:
@2767mr @Vegita2 Change log failure,  is this something I need to commit to the branch?

2767mr (changes requested):
Unfortunately, I don't have a (good) touch screen to test this.

* Missing changelog

2767mr:
> @2767mr @Vegita2 Change log failure, is this something I need to commit to the branch?

Yes, please do.

webapp/src/app/shared/phaser/tilemap/tile-drawer.ts
bawood2: @2767mr Since this is a new feature,  would updated the old shift break users who are used to this feature?  I can update the panKey to `CTRL`, not a deal breaker.  I went with SHIFT initially because it felt more natural.

2767mr: It would not be a big deal because nobody knows it exists 😄

bawood2:
@2767mr Change log updated

webapp/src/app/shared/phaser/tilemap/tile-drawer.ts
bawood2: Sounds good,  I'll push an update

bawood2: Done

bawood2:
@2767mr Do I need to squash these commits before PR goes through?

2767mr:
That won't be not necessary

2767mr (approved):

webapp/src/app/shared/phaser/tilemap/tile-drawer.ts
```diff
@@ -31,6 +31,7 @@ export class TileDrawer extends BaseObject {
 	private fillKey!: Phaser.Input.Keyboard.Key;
 	private lastDraw: Point = {x: -1, y: -1};
 	
+	private panKey!: Phaser.Input.Keyboard.Key;
```
Vegita2, line 34: the pan key is currently defined 3 times in 3 different files. Ideally it should only be defined in map-pan and update some global state so the other functions can access it. E.g. adding a  variable `panning: boolean` to the class `Globals`

Vegita2 (changes requested):
Good addition. I don't use a touchpad so it's good to get some feedback that the editor is actually unusable without a mouse.

A few issues I noticed while testing: 
- a selected entity gets deselected when using the new ctrl panning.
- If you hold the left button down and move the cursor before pressing ctrl, the pointer jumps to the starting position once the ctrl button is pressed.

Other than that it works fine.

webapp/src/app/shared/phaser/tilemap/tile-drawer.ts
bawood2: @Vegita2 Glad you said something,  had the same thought about duplicated code.  I'll give updating this to be globals a shot and push an update when ready.

bawood2:
@Vegita2 

> a selected entity gets deselected when using the new ctrl panning.

Yeah, I tried to find where selected entities were being deselected.  Thought it would be in the `pointerup` key binding,  and tried a few different things, but wasn't able to pin point where the additional logic to prevent deselection would go.  I'll take another look.   However,  holding SHIFT while panning with CTRL keeps the entities selected.

> If you hold the left button down and move the cursor before pressing ctrl, the pointer jumps to the starting position once the ctrl button is pressed.

Hmmm.  I'll take a look

webapp/src/app/shared/phaser/entities/entity-manager.ts
```diff
@@ -126,7 +128,7 @@ export class EntityManager extends BaseObject {
 						entity.startOffset.y = pointer.worldY - entity.container.y;
 						entity.isDragged = true;
 					});
-				} else {
+				} else if (!this.panKey.isDown){
```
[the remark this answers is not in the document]
bawood2: @Vegita2 The reason why I put this here, rather than including this in the return logic where at the begining of the function where `!pointer.leftButtonDown()` is to allow for the capability to multi select while panning.  Upon closer inspection, I do see the entities wiggle when panning,  so I would like to propose adding an `if panKey.isDown then return` statement after the entity has been selected and before the dragging

webapp/src/app/shared/phaser/tilemap/tile-drawer.ts
bawood2: @Vegita2 okay, added a new global, set in `map-pan`,  works good

bawood2:
@Vegita2 

> a selected entity gets deselected when using the new ctrl panning.

Addressed by: https://github.com/CCDirectLink/crosscode-map-editor/pull/207/commits/994c3ecb479493ef961751a280e5ab2baf524775

> If you hold the left button down and move the cursor before pressing ctrl, the pointer jumps to the starting position once the ctrl button is pressed.

Addressed by:  https://github.com/CCDirectLink/crosscode-map-editor/pull/207/commits/e98fa8da9439bb027ad0e26df2141d8d14e3ea80

bawood2:
@2767mr @Vegita2 Comments/Changes have been addressed.  Ready for review

webapp/src/app/shared/phaser/map-pan.ts
```diff
@@ -26,7 +28,7 @@ export class MapPan extends Phaser.GameObjects.GameObject {
 	}
 	
 	private onMouseDown() {
-		if (!this.active) {
+		if (!this.active || !Globals.panning){
```
Vegita2, line 31: I'm pretty sure this line prevents middle mouse panning to work

Vegita2 (changes requested):
I tried it again and now panning with the middle mouse button stopped working. There are also some linting errors. Make sure to run the linting script (`npm run lint`) before pushing changes

webapp/src/app/shared/phaser/map-pan.ts
bawood2: @Vegita2 I'll take a look, and find my mouse to test with that as well

bawood2: @Vegita2 Fixed

bawood2:
@Vegita2 Where all should I run the linter?  Just the web-app?

```
(add_map_pan_key_control) webapp $ npm run lint

> cc-map-editor@0.9.0 lint C:\Users\Nause\crosscode-map-editor\webapp
> ng lint


Linting "map-editor"...
All files pass linting.
```

webapp/src/app/shared/phaser/map-pan.ts
bawood2: Disregard,  this fixed the issue with the mouse wheel,  but created a different issue.  Will ping you again when this is ready

bawood2: @Vegita2 Okay,  fixed fixed

bawood2:
@Vegita2 Comments have been addressed.  Did additional testing with a mouse.  Ready for delta-review

Vegita2 (approved):
> Where all should I run the linter? Just the webapp?

Yes, you only changed code in webapp so it's enough to run just there. Building the app should also show linting errors. We fixed the ci so builds now run on forks. They should also fail on linting errors.

I tried the latest version and most of it works as expected. Some minor issues left, but I don't think it's worth fixing them at all. The code also looks good.

@2767mr Do you have anything else to add?

2767mr (approved):
It now works as one would expect.

I don't really like adding a field to `Globals` that is activily updated but I agree that it's the best solution (without major rewrites).

webapp/src/app/shared/phaser/entities/entity-manager.ts
```diff
@@ -140,6 +145,12 @@ export class EntityManager extends BaseObject {
 					this.selectEntity();
 					this.showAddEntityMenu();
 				} else if (pointer.leftButtonReleased()) {
+
+					// if panning do not deselect entities
```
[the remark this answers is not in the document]
bawood2: @Vegita2 I think this would be an easy fix,  it would just mean putting this logic below the `isDragged = false`,  so that entities stop following the mouse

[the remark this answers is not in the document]
bawood2: @Vegita2 Thanks for providing the context.  Definitely helps understand whats going on and how to modify the behavior.  The place where this should go is right after `const entities = this.selectionBox.onInputUp();`, such that:


-  Selected entities will not be deselected when ctrl is released, and then ctrl is used for more panning
-  Pressing ctrl while dragging entities then releasing mouse will not cause entities to follow mouse when additional panning is performed.
entities won't follow the mouse when dragged, and the selection box will be removed.
-  Pressing ctrl in the middle of a selection box will remove the selection box once the mouse button is released.

Even though this has been approved, would like to make this modification, since its simple.

webapp/src/app/shared/phaser/map-pan.ts
```diff
@@ -69,8 +80,16 @@ export class MapPan extends Phaser.GameObjects.GameObject {
 	}
 	
 	preUpdate() {
+
+		if (this.panKey.isDown || this.scene.input.activePointer.middleButtonDown()) {
```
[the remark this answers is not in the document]
bawood2: @Vegita2 Thanks for the feedback.  Knew that assignments like this were allowed in python,  didn't realize they were allowed in javascript.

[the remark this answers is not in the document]
bawood2: Updated

webapp/src/app/shared/phaser/entities/entity-manager.ts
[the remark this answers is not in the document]
bawood2: Updated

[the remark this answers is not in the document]
Vegita2: The problem with this is that the selection box now disappears, but the entities it selected are ignored. That means they are still in the hovered over state, but not actually selected. The hover state disappears when hovering over them again.

webapp/src/app/shared/phaser/map-pan.ts
```diff
@@ -81,11 +81,11 @@ export class MapPan extends Phaser.GameObjects.GameObject {
...
 	preUpdate() {
 
-		Globals.panning = this.panKey.isDown || this.scene.input.activePointer.middleButtonDown();
-
 		const pointer = this.scene.input.activePointer;
+		
+		Globals.panning = this.panKey.isDown || pointer.middleButtonDown();
 
-		if (Globals.panning && this.isScrolling && this.active) {
+		if (Globals.panning && this.isScrolling && this.active && pointer.isDown) {
```
bawood2, line 88: @Vegita2 Found a case where onMouseUp wasn't being called when the pointer was off the screen.  When this occurs, `isScrolling` doesn't ever get updated.  Pressing ctrl in this case would move the entire camera, even though scrolling is not active.  Added `pointer.isDown` to prevent this from happening.

webapp/src/app/shared/phaser/entities/entity-manager.ts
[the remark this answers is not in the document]
bawood2: Good catch.  Let me take another look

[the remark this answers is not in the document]
bawood2: @Vegita2 Tried a few different things, but can't seem to get it to work.  Want to avoid modifying the `selectEntity` method until I have a better understanding of what its doing, especially since its used in multiple places.  For now,  moved the return until after `isDragged = false`.  This brings back the issue where the selection box persists if ctrl is is pressed during the selection box.  How do you want to capture this defect,  should I post an issue on it?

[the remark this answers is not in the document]
Vegita2: As i said i don't think this issue matters to much and we should just leave it as is. I don't think anyone ever tried to pan the map while simultaneously selecting entities.

PR #212, opened 2021-03-02
Add Enemy support

2767mr:
This PR adds support for rendering enemies as well as widgets for editing their properties.

Things that might need to be discussed:
* Boss (multi-part-) entities are broken but I cannot find out why. I decided to leave it in a semi-broken state because bosses are not common and do not need to be rendered perfectly in the first place.
* The  enemy widget has no clear way of showing where the enemy data begins and where it ends. This is mostly due to a lack of space. Any suggestions for improvements appreciated.
* Dropdowns do not seem to have the correct style. I left it as is because they are used in other places of the editor.


Vegita2:
> The enemy widget has no clear way of showing where the enemy data begins and where it ends. This is mostly due to a lack of space. Any suggestions for improvements appreciated.

maybe something similar to how the crosscode devs did it? Basically showing another window similar to how npc states and events work. 

image

webapp/src/app/models/multi-entity-animation.ts
```diff
@@ -0,0 +1,57 @@
+import { Point3 } from './cross-code-map';
+import { TileSheet } from './tile-sheet';
+
+export interface MultiEntityAnimation {
+    DOCTYPE: 'MULTI_ENTITY_ANIMATION';
+
+    anims: Record<string, EntityAnim>;
+    baseSize: Point3;
+    namedSheets?: Record<string, TileSheet>;
```
Vegita2, line 9: isn't this type just `{ [key: string]: TileSheet }`?

webapp/src/styles.scss
```diff
@@ -61,6 +61,26 @@ input[type=text].default-input, input:not([type]).default-input {
 	outline: solid 1px mat-color($mat-blue, 500);
 }
 
+
+.default-input.mat-select {
```
Vegita2, line 65: Where are these styles used?

Vegita2 (changes requested):
Looks good.

Only thing is the added styles in `styles.scss`.
I'm not sure if they are still used or not

webapp/src/app/models/multi-entity-animation.ts
2767mr: That is correct - `type Record<A, B> = { [key: A]: B }`. Should I change it?

webapp/src/styles.scss
2767mr: They are used for selects inside widgets. Without these there are no borders and the size is slightly messed up. See https://github.com/CCDirectLink/crosscode-map-editor/blob/enemy/webapp/src/app/shared/widgets/string-widget/string-widget.component.html#L4-L10

webapp/src/app/models/multi-entity-animation.ts
Vegita2: I don't really care what we use. I just never saw `Record`used before

webapp/src/styles.scss
Vegita2: alright

Vegita2 (approved):

PR #217, opened 2021-05-21
small issues

Vegita2:
fixes some smaller issues, checkout changelog/commits for more details,


webapp/src/app/shared/phaser/main-scene.ts
```diff
@@ -8,49 +8,55 @@ import { TileDrawer } from './tilemap/tile-drawer';
...
 		this.sub = Globals.mapLoaderService.map.subscribe((map) => {
 			if (map) {
 				tileMap.loadMap(map);
 				this.rescaleBorder();
+				
+				// reset camera position on map load
+				const cam = this.cameras.main;
+				cam.scrollX = 0;
+				cam.scrollY = 0;
+				cam.zoom = 1;
```
2767mr, line 59: Possible improvement: center map in center of viewport instead of top left

webapp/src/app/shared/phaser/map-pan.ts
```diff
@@ -61,7 +61,7 @@ export class MapPan extends Phaser.GameObjects.GameObject {
 		
 		let zoom = delta > 0 ? 0.8 : 1.25;
 		zoom *= cam.zoom;
-		if (zoom > 0.4 && zoom < 8) {
+		if (zoom > 0.4 && zoom < 50) {
```
2767mr, line 64: This change is not mentioned in the change log and I don't see any particular reason for it. Was the change intentional?

2767mr (approved):

webapp/src/app/shared/widgets/event-widget/event-editor/editor/event-editor.component.ts
```diff
@@ -64,17 +65,25 @@ export class EventEditorComponent implements OnChanges {
...
-
+	
+	ngOnDestroy() {
+		EventEditorComponent.globalBase = this.base;
+	}
+	
+	show() {
+		this.detailsShown = false;
+	}
+
```
2767mr, line 76: Maybe use getters/setters instead

```suggestion
	get base() {
		return EventEditorComponent.globalBase;
	}

	set base(value: number) {
		EventEditorComponent.globalBase = value;
	}
```

webapp/src/app/shared/phaser/map-pan.ts
Vegita2: forgot to add it in the changelog. It's for high res displays. The editor tries to display everything pixel perfect, so with a max zoom of 8 it is roughly comparable to running crosscode on a 4k monitor at full screen, not actually zooming into the map.

webapp/src/app/shared/phaser/main-scene.ts
Vegita2: implemented it. 
It required changing the dimensions of the containing canvas element to accurately get the center. Could cause issues on different display resolutions, so it would be good if you could run it and check if the yellow borders are visible on the bottom or right like in that image.

image

webapp/src/app/shared/widgets/event-widget/event-editor/editor/event-editor.component.ts
Vegita2: implemented

PR #219, opened 2021-09-20
Add setting to wrap long event editor lines over multiple lines

Image-unavailable:
Adds a checkbox in the settings that allows to switch the wrapping behaviour of long lines in the event editor between clipping and wrapping them over multiple lines.

Checkbox:
Checkbox screenshot

Comparison between the two modes:
Mode comparison

### DIFF
--- a/webapp/src/app/services/sharedService.ts
+++ b/webapp/src/app/services/sharedService.ts
@@ -1,6 +1,9 @@
 export interface SharedService {
     saveModSelect(mod: string): Promise<void>;
     getSelectedMod(): string;
+	
+	saveWrap (wrap: boolean): void;
+	getWrap (): boolean;
     
     relaunch(): void;
 }
--- a/webapp/src/app/services/sharedService.ts
+++ b/webapp/src/app/services/sharedService.ts
@@ -2,8 +2,8 @@ export interface SharedService {
     saveModSelect(mod: string): Promise<void>;
     getSelectedMod(): string;
 	
-	saveWrap (wrap: boolean): void;
-	getWrap (): boolean;
+	saveWrapEventEditorLinesSetting(wrap: boolean): void;
+	getWrapEventEditorLinesSetting(): boolean;
     
     relaunch(): void;
 }
--- a/webapp/src/app/services/sharedService.ts
+++ b/webapp/src/app/services/sharedService.ts
@@ -2,8 +2,5 @@ export interface SharedService {
     saveModSelect(mod: string): Promise<void>;
     getSelectedMod(): string;
 	
-	saveWrapEventEditorLinesSetting(wrap: boolean): void;
-	getWrapEventEditorLinesSetting(): boolean;
-    
     relaunch(): void;
 }
--- a/webapp/src/app/shared/settings.service.ts
+++ b/webapp/src/app/shared/settings.service.ts
@@ -1,0 +1,18 @@
+import {Injectable} from '@angular/core';
+
+@Injectable()
+export class SettingsService {
+	private static readonly wrapSettingName = 'wrapEventEditorLines';
+	
+	private static loadBooleanOrDefault(key: string, defaultValue: boolean): boolean {
+		const loadedValue = localStorage.getItem(SettingsService.wrapSettingName);
+		return loadedValue === null? defaultValue : (loadedValue === 'true');
+	}
+	
+	get wrapEventEditorLines() {
+		return SettingsService.loadBooleanOrDefault(SettingsService.wrapSettingName, true);
+	}
+	set wrapEventEditorLines(value: boolean) {
+		localStorage.setItem(SettingsService.wrapSettingName, value.toString());
+	}
+}
--- a/webapp/src/app/shared/widgets/event-widget/event-editor/editor/event-editor.component.ts
+++ b/webapp/src/app/shared/widgets/event-widget/event-editor/editor/event-editor.component.ts
@@ -38,12 +38,10 @@ export class EventEditorComponent implements OnChanges {
 		EventEditorComponent.globalBase = value;
 	}
 	
-	get wrapText (): boolean {
-		const result = this.wrapTextOverride === undefined?
-			this.sharedService.getWrap () :
+	get wrapText(): boolean {
+		return this.wrapTextOverride === undefined?
+			this.sharedService.getWrapEventEditorLinesSetting() :
 			this.wrapTextOverride;
-		
-		return result;
 	}
 	
 	detailsShown = false;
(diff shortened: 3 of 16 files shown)


webapp/src/app/shared/widgets/event-widget/event-editor/editor/event-editor.component.ts
```diff
@@ -25,6 +29,7 @@ export class EventEditorComponent implements OnChanges {
 	
 	@Input() eventData: EventType[] = [];
 	@Input() actionStep = false;
+	@Input() wrapTextOverride?: boolean; //If assigned uses the assigned value, if undefined uses the value from the settings
```
Vegita2, line 32: I would remove this currently because it's not used. If there is ever a use case for it it's not difficult to implement.

webapp/src/app/shared/widgets/event-widget/event-editor/editor/event-editor.component.ts
```diff
@@ -33,6 +38,12 @@ export class EventEditorComponent implements OnChanges {
 		EventEditorComponent.globalBase = value;
 	}
 	
+	get wrapText(): boolean {
+		return this.wrapTextOverride === undefined?
```
Vegita2, line 42: can be simplified to 
```typescript
get wrapText(): boolean {
    return this.wrapTextOverride ?? this.sharedService.getWrapEventEditorLinesSetting();
}
```

Also try to avoid calling functions in the template because they get called a billion times (every time angular runs change detection).
A better option would be to calculate the value in `ngOnInit` and cache it or if the function has an input parameter use a pipe instead.

Vegita2 (changes requested):
Looks good so far,

One thing I don't really like is the position of the circle on wrapped lines, I would prefer something like this one where it is one the first line instead of the center. What do you think?
image

Also changelog entry is missing.

webapp/src/app/services/sharedService.ts
```diff
@@ -1,6 +1,9 @@
 export interface SharedService {
     saveModSelect(mod: string): Promise<void>;
     getSelectedMod(): string;
+	
+	saveWrapEventEditorLinesSetting(wrap: boolean): void;
```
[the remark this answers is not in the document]
Image-unavailable: I used `SharedService` because because that's what the other settings used, so I just thought that's how settings persistance worked.
As for where the wrap setting should be I feel like the best way to do it is probably to have something like a `SettingsService` instead of putting it in `GlobalEventsService`, that way we don't mix persistent settings with other things and there is already a place for eventual future settings.

Image-unavailable:
> One thing I don't really like is the position of the circle on wrapped lines, I would prefer something like this one where it is one the first line instead of the center. What do you think?

I think it's better than centered dot, it's more practical and it feels more consistent, skimming events also feels way better since having the dot always at the start of the line makes it easy to tell them apart, unlike with centered dot where it feels like it's more of a distraction than anything else (I didn't think about dot placement at all when I made the changes to the CSS...).

Image-unavailable:
Also a minor problem with the color preview of `SET_OVERLAY`, the colored area gets split over two lines with some editor widths:
immagine
immagine
(image from ms. Teak in west rookie harbor, npc state page 6)

webapp/src/app/services/sharedService.ts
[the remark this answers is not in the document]
Vegita2: sounds good

Image-unavailable:
The colored area splitting over multiple lines can be fixed easily by replacing normal spaces in the color rectangle with non-breaking spaces.
Not sure whether it looks better if it's split or not, but a single rectangle is probably what most people would expect it to look like, so maybe it's the better option?
immagine
immagine

webapp/src/app/shared/settings.service.ts
```diff
@@ -0,0 +1,18 @@
...
+	
+	private static loadBooleanOrDefault(key: string, defaultValue: boolean): boolean {
+		const loadedValue = localStorage.getItem(SettingsService.wrapSettingName);
+		return loadedValue === null? defaultValue : (loadedValue === 'true');
+	}
+	
+	get wrapEventEditorLines() {
+		return SettingsService.loadBooleanOrDefault(SettingsService.wrapSettingName, true);
+	}
+	set wrapEventEditorLines(value: boolean) {
```
Vegita2, line 15: I would personally make all settings into observables, so they can be changed on the fly and can be safely used inside html templates without functions getting called all the time.

So basically something like this: 
```typescript
@Injectable()
export class SettingsService {
	private static readonly wrapSettingName = 'wrapEventEditorLines';
	
	private wrapEditorLines = new BehaviorSubject(
		SettingsService.loadBooleanOrDefault(SettingsService.wrapSettingName, true)
	);
	
	private static loadBooleanOrDefault(key: string, defaultValue: boolean): boolean {
		const loadedValue = localStorage.getItem(SettingsService.wrapSettingName);
		return loadedValue === null ? defaultValue : (loadedValue === 'true');
	}
	
	getWrapEventEditorLines() {
		return this.wrapEditorLines.asObservable();
	}
	
	setWrapEventEditorLines(val: boolean) {
		localStorage.setItem(SettingsService.wrapSettingName, val.toString());
		this.wrapEditorLines.next(val);
	}
}
```

Now in the event editor this can be used like that:

```typescript
export class EventEditorComponent implements OnChanges, OnInit {
	[...]
	$wrapText: Observable<boolean>;

	constructor(
		private helper: EventHelperService,
		private addEvent: AddEventService,
		settingsService: SettingsService
	) {
		this.$wrapText = settingsService.getWrapEventEditorLines();
	}
	[...]
```

```html
<div class="event-container" [ngClass]="{'wrapping': $wrapText | async}" fxLayout="row">
	[...]
```

The benefit of this is that the option can be changed at any time and is immediately reflected in the ui without reloading the component or calling a get function over and over again.

That said it's good how it is. There is no way to change the option without reloading the component anyway. Just for future reference.

Vegita2 (approved):
It's definitely better to show a single rectangle instead of splitting it over multiple lines. 
Ideally it would keep all `key:value` pairs together without wrapping, but I don't think it's necessary and just fixing the color block wrapping is enough.

Other than that it looks good. Once you push your changes to fix the color block we can merge it and release a new version

Vegita2:
I agree it looks better. Can you add that too?

Image-unavailable:
I also added word breaks to `SHOW_MSG`, `ADD_MSG_PERSON`, `SHOW_CHOICE` so that long names and expressions get displayed better:
immagine

Image-unavailable:
I also added word breaks to `SHOW_MSG`, `ADD_MSG_PERSON` and `SHOW_CHOICE` so that long names and expressions get displayed better:

(no word break)
immagine
(with word break)
immagine

Vegita2:
Looks good. I'll try it out later today and if I don't see anything wrong I'll merge it.

Vegita2 (changes requested):
`show-msg` and the other components that use `<wbr>` don't behave correctly when line breaks are disabled. They still add a line break.

image

Vegita2 (approved):

Issue #188, opened 2020-10-04
Ability to mark NPC Pages Event type 


inkkly:
Add the ability to mark the event type (identified in the source code as `sc.NPC_EVENT_TYPE`) by nesting that pages event and action steps inside their respective identifiers:

## `SIMPLE`
No change, since this is the default :p 

## `TRADE`
More complicated and nuanced than the other types.
Nests `"trade":{"trader": "nameofTrader"}` inside of the event, with `nameofTrader` being set to whatever the internal name of the trader is in `database.json`:

```
"event": {
    "trade": {
        "trader": "rhombusPostHead1"
    }
}
```

It's important to note that event steps are not typically used with trade NPCs, except in rare cases like the Geta Trader, where they're usually done by nesting another `event` array inside of `trade`, and then using the `traded` and `canceled` branches of `START_NPC_TRADE_MENU`: 
```javascript
"event" :{
    "trade": {
        "event": [{
            "withBranches": true,
            "type": "START_NPC_TRADE_MENU",
            "traded": [ /*events go in here*/ ],
            "canceled": [ /*or here*/ ]
        }], 
    "trader": "sapphireGeta"
    }
}
```

Though, you could feasibly also add event steps inside of event>trade>event if you want to have a dialogue before opening the trade menu, just make sure START_NPC_TRADE_MENU is an event step so the trade still actually happens.

## `QUEST`, `SHOP`, and `ARENA`
Much simpler compared to `trade`, they simply add `quest` or `arena` arrays inside of the event, inside of which event steps are added:

```javascript
"event": {
        "quest": [
               //events go here
          ]
}
```

2767mr:
Closed by #221

PR #221, opened 2021-10-28
Solve issue 188

Image-unavailable:
Solves #188.
Add the ability to load `shop`, `arena` and `trade` events in the event editor and save them correctly (including `quest` events).
The event type can be specified in the side panel of NPC editing windows:
Side panel

Trade events add another property below the event type to specify the trader name:
Trader name field

Trade events have a warning that pops up when a trader is not specified:
Missing trader name warning

There is also a warning if a trade event does not have a `START_NPC_TRADE_MENU` event:
Missing start trade event warning
The warning does not apper on traders with no events since the `START_NPC_TRADE_MENU` is not required if the trader has its `event` property not set and the editor exports traders with an empty event array as undefined.

### DIFF
--- a/webapp/src/app/shared/widgets/event-widget/event-editor/editor/event-editor.component.ts
+++ b/webapp/src/app/shared/widgets/event-widget/event-editor/editor/event-editor.component.ts
@@ -72,13 +72,9 @@ export class EventEditorComponent implements OnChanges, OnInit {
 	ngOnChanges() {
 		const eventCopy: EventArray = JSON.parse(JSON.stringify(this.eventData));
 		const destructuredEvents = destructureEventArray(eventCopy);
-		if (destructuredEvents !== undefined) {
-			this.inputtedEventType = destructuredEvents.type;
-		} else {
-			console.error('Loaded map has invalid format.');
-		}
 		
-		this.workingData = destructuredEvents.events.map((val: EventType) => this.helper.getEventFromType(val, this.actionStep));
+		this.inputtedEventType = destructuredEvents.type;
+		this.workingData = destructuredEvents.events?.map((val: EventType) => this.helper.getEventFromType(val, this.actionStep)) ?? [];
 		
 		this.refreshAll();
 	}
--- a/webapp/src/app/shared/widgets/event-widget/event-editor/editor/event-editor.component.ts
+++ b/webapp/src/app/shared/widgets/event-widget/event-editor/editor/event-editor.component.ts
@@ -28,7 +28,7 @@ export class EventEditorComponent implements OnChanges, OnInit {
 	@Input() eventData: EventArray | unknown = [];
 	@Input() actionStep = false;
 	
-	@Output() eventsChanged = new EventEmitter<EventArray>();
+	@Output() eventsChanged = new EventEmitter<EventType[]>();
 	
 	get base() {
 		return EventEditorComponent.globalBase;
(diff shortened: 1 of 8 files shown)


webapp/src/app/shared/widgets/npc-states-widget/npc-states/npc-states.component.ts
```diff
@@ -42,19 +47,27 @@ export class NpcStatesComponent implements OnInit {
 		this.selectTab(this.index);
 	}
 	
+	ngDoCheck() {
```
Vegita2, line 50: Don't use this function without a very good reason. It's only there to manually detect changes and gets called way too often.
Recalculating warnings can be triggered manually instead

webapp/src/app/shared/widgets/npc-states-widget/npc-states/npc-states.component.ts
```diff
@@ -42,19 +47,27 @@ export class NpcStatesComponent implements OnInit {
...
+		this.warnings.length = 0;
+		if (this.isTradeEvent && !this.trader) {
+			this.warnings.push('Missing trader name');
+		}
+		if (this.eventType === EventArrayType.Trade && this.missingTradeEvent) {
```
Vegita2, line 55: `this.isTradeEvent` does the same thing, could be in the same if block

webapp/src/app/shared/widgets/npc-states-widget/npc-states/npc-states.component.ts
```diff
@@ -125,4 +138,30 @@ export class NpcStatesComponent implements OnInit {
...
+		this.missingTradeEvent =
+			updatedEvents.length > 0 &&
+			!hasEventOfTypeRecursive(updatedEvents, 'START_NPC_TRADE_MENU');
+	}
+	
+	get isTradeEvent() {
+		return this.eventType === EventArrayType.Trade;
+	}
+	
+	get eventTypes(): string[] {
```
Vegita2, line 164: should be calculated only once and reused. functions in templates are really inefficient

webapp/src/app/shared/widgets/npc-states-widget/npc-states/npc-states.component.ts
```diff
@@ -125,4 +138,30 @@ export class NpcStatesComponent implements OnInit {
...
+		return Object.keys(EventArrayType).map(name => (EventArrayType as any)[name]);
```
Vegita2, line 165: can be simplified
```typescript
return Object.values(EventArrayType);
```

webapp/src/app/shared/widgets/event-widget/event-editor/editor/event-editor.component.ts
```diff
@@ -103,8 +113,12 @@ export class EventEditorComponent implements OnChanges, OnInit {
 		this.detailsShown = false;
 	}
 	
-	export() {
-		return this.workingData?.map(event => event.export()) ?? this.eventData;
+	export(eventType?: EventArrayType, trader?: string): EventArray {
```
Vegita2, line 116: Why is `EventArrayType` handled in the `EventEditorComponent` instead of the `NpcStatesComponent`? My understanding is that quest/trader etc is only relevant for NPC's so they should take care of properly saving it.

Vegita2 (changes requested):
UI looks good. Code is also fine with a few minor issues. We can merge it after you fix them.

Vegita2 (approved):
Looks good. Do you want to add/change anything or can we merge it?

Image-unavailable:
I don't have any other changes, you can merge it.

PR #229, opened 2022-04-25
Implement coordinate display

### LINKED ISSUE #227: Show coordinates of mouse cursor
A lot of action steps, for example `MOVE_TO_POINT`, take a set of coordinates to move an entity to.

However... It's not really clear how you're supposed to get your desired coordinates.

Sure, you can create a dummy entity somewhere, extract the coordinates, and delete it, but that's dirty.

It would be handy if the coordinates of, say, your mouse position in the map would be displayed somewhere.

yellowsink:
Closes #227, see that issue for more details.

Shows both the coords of the tile under the cursor, and the exact cursor coords for entities.

Image:


yellowsink:
CI passed on my fork, waiting on run here :)

yellowsink:
so it appears that npm is completely broken on the windows CI? possibly an affect of updating the npm locks from an old format?

image

webapp/src/app/shared/phaser/coords-reporter.ts
```diff
@@ -0,0 +1,65 @@
...
+
+		this.rawCoords = newRaw;
+		this.lastModeWasTile = tileMode;
+
+		// send event
+		Globals.globalEventsService.updateCoords.next({
+			...this.coords,
+			z: this.entityLevel,
+		});
+	}
```
Vegita2, line 64: subscribe to subjects instead of polling the view every frame. The preUpdate function should always report the coordinates. If the view changes the actual game object can be disabled.
Something like this:

```typescript
import { Point } from "../../models/cross-code-map";
import { EditorView } from "../../models/editor-view";
import { Globals } from "../globals";
import { Subscription } from "rxjs";
import { BaseObject } from "./base-object";

export class CoordsReporter extends BaseObject {
  private rawCoords: Point = { x: 0, y: 0 };

  private sub!: Subscription;

  constructor(scene: Phaser.Scene, active = true) {
    super(scene, "coordsReporter", active);
  }

  protected init(): void {
    this.sub = Globals.globalEventsService.currentView.subscribe((view) => {
      this.setActive(view === EditorView.Entities);
    });
  }

  protected activate(): void {}

  protected deactivate(): void {
    Globals.globalEventsService.updateCoords.next();
  }

  destroy(fromScene?: boolean) {
    this.sub.unsubscribe();
    super.destroy(fromScene);
  }

  private get coords(): Point {
    return {
      x: this.rawCoords.x,
      y: this.rawCoords.y + this.entityLevel,
    };
  }

  private get entityLevel() {
    const maybeLevel =
      Globals.globalEventsService.selectedEntity.value?.details.level;

    return maybeLevel
      ? Globals.map.levels[maybeLevel.level].height + maybeLevel.offset
      : 0;
  }

  preUpdate(time: number, delta: number) {
    const pointer = this.scene.input.activePointer;

    const newRaw = {
      x: Math.floor(pointer.worldX),
      y: Math.floor(pointer.worldY),
    };

    // don't send events if we don't have to
    if (newRaw.x === this.rawCoords.x && newRaw.y === this.rawCoords.y) {
      return;
    }

    this.rawCoords = newRaw;

    // send event
    Globals.globalEventsService.updateCoords.next({
      ...this.coords,
      z: this.entityLevel,
    });
  }
}

```
I also added the `BaseObject` class to have proper typings for preUpdate

webapp/src/app/shared/phaser/coords-reporter.ts
```diff
@@ -0,0 +1,65 @@
...
+
+		const newRaw = {
+			x: Math.floor(pointer.worldX),
+			y: Math.floor(pointer.worldY),
+		};
+
+		// don't send events if we don't have to
+		if (newRaw.x === this.rawCoords.x && newRaw.y === this.rawCoords.y) {
+			return;
+		}
```
Vegita2, line 54: this condition ignores changes to `z`. Selecting an entity without moving the mouse doesn't update the view

webapp/src/app/components/toolbar/toolbar.component.ts
```diff
@@ -72,6 +73,10 @@ export class ToolbarComponent implements OnInit {
 		}
 
 		this.events.babylonLoading.subscribe(val => this.is3dLoading = val);
+
+		this.events.updateCoords.subscribe(coords => {
+			this.coords = coords === null ? '' : `(${coords.x}, ${coords.y}, ${coords.z})`;
```
Vegita2, line 78: This could also be rewritten to 

```suggestion
			this.coords = coords ? `(${coords.x}, ${coords.y}, ${coords.z})` : '';
```

which makes it a bit easier to read in my opinion

webapp/src/app/shared/global-events.service.ts
```diff
@@ -17,6 +17,8 @@ export class GlobalEventsService {
 	offsetMap = new Subject<Point>();
 	toggleVisibility = new Subject<void>();
 	showAddEntityMenu = new Subject<Point>();
+
+  updateCoords = new Subject<Point3 | null>();
```
Vegita2, line 21: use `undefined` instead of `null`, also wrong indentation
```suggestion
	updateCoords = new Subject<Point3 | undefined>();
```

webapp/src/app/shared/phaser/coords-reporter.ts
```diff
@@ -0,0 +1,65 @@
...
+			x: this.rawCoords.x,
+			y: this.rawCoords.y + this.entityLevel,
+		};
+	}
+
+	private get isEntityMode() {
+		return (
+			Globals.globalEventsService.currentView.value ===
+			EditorView.Entities
+		);
```
Vegita2, line 19: no need for parentheses and line breaks. It just makes it less readable 
```suggestion
		return Globals.globalEventsService.currentView.value === EditorView.Entities;
```

Vegita2 (changes requested):
Looks good so far,

The changing `y` coordinate based on entity height confused me a bit, but it's probably better to display the sum of `y` and `z` because that is what you normally want.
Also why are the coordinates only showing in entity view? 

Added a few suggestions. 
Mainly the `CoordsReporter` should subscribe to the currentview event instead of polling them and moving the overlays into a separate component (coordinates and version).

There are also some indentation issues. You can use the provided .editorconfig to have your editor automatically fix them

yellowsink:
Much of the formatting issues are due to Prettier, will fix.

The reason they only show in entity view was due to discussion over the display in the Discord channel - it isn't massively useful outside of entity positioning and also requires additional logic to correctly resolve Z info for both layer and entity view

webapp/src/app/shared/phaser/coords-reporter.ts
yellowsink: I see, nice to know where that bug comes from.

This check I don't actually think it strictly necessary, but sending useless events seemed like something to filter out.

yellowsink:
I'm assuming its okay for me to add a prettier config file that just tells it to follow the editorconfig and to play nice with eslint (specifically, singlequote)
image

webapp/src/app/components/toolbar/toolbar.component.ts
yellowsink: I've done `!coords ? '' : '...'` but its basically the same thing

yellowsink:
whyd it work locally then bruh
image

Vegita2 (approved):
Looks good now. Added a small change to disable coords display in 3d

yellowsink:
:tada:

Issue #159, opened 2020-05-04
support custom input for all settings (mods)


Vegita2:
request from discord:

If you're using the map editor casually, these drop downs on settings are really helpful for finding areas/music/tilesets you need. But there is no custom feature. There's no way to input anything through this dropdown system.

Please dont just.. Remove the drop down. Just add an "other" or "custom" fill out section for mod makers.
Please and thank you :GWllentNicoPeek: :sparkling_heart:
![image](https://user-images.githubusercontent.com/9483499/80986801-5ba05f80-8e31-11ea-84c7-52f7e5eb354d.png)

![image](https://user-images.githubusercontent.com/9483499/80986827-64913100-8e31-11ea-8fe7-d21b1e8d611d.png)

PR #233, opened 2022-08-16
Fix step definitions, add autocompletion to `String` inputs (issue #232)

### LINKED ISSUE #232: Bug: Party member event steps can't accept strings
The map editor's event editor does not accept strings for the **member** attribute for the following event steps:
ADD_PARTY_MEMBER
REMOVE_PARTY_MEMBER
SET_PARTY_MEMBER_ALL_ELEMENTS
SET_PARTY_MEMBER_LEVEL
SET_PARTY_MEMBER_NO_DIE
SET_PARTY_MEMBER_SP_LEVEL

Currently, the proper strings need to be defined in the map's .json file since the map editor only seems to allow numbers.

image

Image-unavailable:
Fixes inability to edit the `member` property of certain party member-related event steps. (Fixes #232)
This PR also fixes `SET_MEMBER_LOCKED`, which was not specified in the issue.

Before:
immagine

After:
immagine


2767mr (changes requested):
`events.json` and `actions.json` are actually created by `tools/parse-action-step.js` (copy it into the game console and it will put the result in your clipboard)
Please fix that script instead, it seems that there are also some instances in `actions.json` that would be fixed.

webapp/src/app/shared/widgets/string-widget/string-widget.component.html
```diff
@@ -1,13 +1,17 @@
...
-				{{ option }}
+		<input class="default-input"
+			[id]="key"
+			[matAutocomplete]="auto"
+			[ngModel]="settings[key]"
+			(ngModelChange)="setSetting(key, $event); updateSuggestedOptions()"
+		>
+		<mat-autocomplete #auto="matAutocomplete" class="dark-scrollbar">
+			<mat-option *ngFor="let option of suggestedOptions" [value]="option">
+				<span appHighlight [highlightText]="option" [highlightMatch]="settings[key]"></span>
```
2767mr, line 12: Since the filter logic now ignores `-`and `_` it would be nice if `appHighlight` correctly colors those.

2767mr (changes requested):
This change looks very good already. I found one small error and a lot of nitpicky stuff.

webapp/src/app/shared/widgets/string-widget/string-widget.component.html
```diff
@@ -1,13 +1,17 @@
...
-		            (selectionChange)="setSetting(key, $event.value)">
-			<mat-option *ngFor="let option of keys" [value]="option">
-				{{ option }}
+		<input class="default-input"
+			[id]="key"
+			[matAutocomplete]="auto"
+			[ngModel]="settings[key]"
+			(ngModelChange)="setSetting(key, $event); updateSuggestedOptions()"
+		>
+		<mat-autocomplete #auto="matAutocomplete" class="dark-scrollbar">
```
2767mr, line 10: This option would be nice
```suggestion
		<mat-autocomplete #auto="matAutocomplete" class="dark-scrollbar" [autoActiveFirstOption]="true">
```

webapp/src/app/shared/widgets/string-widget/string-widget.component.ts
```diff
@@ -23,5 +28,34 @@ export class StringWidgetComponent extends AbstractWidget implements OnInit {
...
 			}
 		}
+		this.updateSuggestedOptions();
+	}
+	
+	updateSuggestedOptions() {
+		const inputText = this.settings[this.key];
+		const searchResults = this.searchFilterService.filterOptions(this.keys, inputText);
+		
+		if (searchResults.some(option => option.length === inputText.length)) {
```
2767mr, line 38: `inputText` may be null or undefined. I would rather have an exact check or a `.startsWith` here anyway since it can lead to confusing ordering otherwise.

2767mr (changes requested):
If you enter `_` it will change suggestions to include `undefined`.

The highlighting looks really nice now.

2767mr (commented):
If you enter regex like `()` it will also result in strange suggestions

2767mr (approved):

webapp/src/app/shared/widgets/string-widget/string-widget.component.ts
```diff
@@ -23,5 +28,34 @@ export class StringWidgetComponent extends AbstractWidget implements OnInit {
...
+				}
+			}
+			this.pushNoDupe(this.suggestedOptions, ...searchResults);
+			this.pushNoDupe(this.suggestedOptions, ...this.keys);
+		} else {
+			this.suggestedOptions = searchResults;
+		}
+	}
+	
+	private pushNoDupe<T>(target: T[], ...values: T[]) {
```
Vegita2, line 54: Why aren't you using a `Set`? It would handle the duplicates by itself

Vegita2 (changes requested):
Looks really good so far.
Adding this new autocomplete to the map settings and tileset selector would also fix #159

What I would like to see is some kind of warning if the value is not present in the options. Something like this:
image

Another thing I noticed is the autocomplete field doesn't scale like the other ones

image

Other than that it works great. Especially the highlighting now works properly

webapp/src/app/shared/widgets/string-widget/string-widget.component.ts
```diff
@@ -23,5 +28,34 @@ export class StringWidgetComponent extends AbstractWidget implements OnInit {
...
+			//This makes it so that if the text is the same as one of the search results all search results are shown (emulates selector-like behaviour).
+			//Matching values are always shown before all the other ones.
+			this.suggestedOptions.length = 0;
+			for (const searchResult of searchResults) {
+				if (searchResult.length === inputText.length) {
+					this.suggestedOptions.push(searchResult);
+				}
+			}
+			this.pushNoDupe(this.suggestedOptions, ...searchResults);
+			this.pushNoDupe(this.suggestedOptions, ...this.keys);
```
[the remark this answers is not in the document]
Image-unavailable: What is it you don't like about it? Autocompletion clutter?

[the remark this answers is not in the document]
Vegita2: I generally expect autocomplete to filter based on the input. E.g. jump height has possible values S, M, L, XL, XXL. So if you now type XL it still shows every value where I would only expect XL and XXL to show.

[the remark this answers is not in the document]
Vegita2: I'm fine with keeping it how it is currently. Just wanted to mention it.

PR #235, opened 2022-08-16
Add widget for editing `langlabel`s

Image-unavailable:
Add a widget for editing `langlabel`s, replacing the current JSON editor.

Current editor:
immagine
immagine

Widget preview:
immagine

https://user-images.githubusercontent.com/56268917/184977793-24777125-6229-466d-beb0-174aa8466a7b.mp4


webapp/src/app/shared/widgets/langlabel-widget/langlabel-widget.component.html
```diff
@@ -0,0 +1,41 @@
+<div fxLayout="row" class="property-container" [matTooltip]="attribute?.description">
+	<label [for]="key" class="property-name">{{key}}:</label>
```
Vegita2, line 2: there is not element with the id `key` set, so the `for` attribute shouldn't be set here

webapp/src/app/shared/widgets/langlabel-widget/langlabel-widget.component.ts
```diff
@@ -0,0 +1,41 @@
...
+		'de_DE',
+		'ja_JP',
+		'ko_KR',
+		'zh_CN',
+		'zh_TW'
+	];
+	
+	constructor() {
+		super();
+	}
```
Vegita2, line 24: empty constructor can be omitted

webapp/src/app/shared/widgets/langlabel-widget/langlabel-widget.component.ts
```diff
@@ -0,0 +1,41 @@
...
+	
+	get hasValue(): boolean {
+		return this.settings[this.key] !== undefined;
+	}
+	
+	get defaultValue() {
+		const value = Object.create(null);
```
Vegita2, line 31: why not `value = {}`? Is there any difference?

Vegita2 (approved):
Looks good,
the only thing I would change is to keep the value when toggling `hasValue` off and on again, but that's not really important.

webapp/src/app/shared/widgets/langlabel-widget/langlabel-widget.component.ts
Image-unavailable: The difference between `Object.create(null)` and `{}` is that if (for some reason) the object's prototype has been modified, the fields that have been added show up when iterating the object with a `for in` loop an object created with `{}`, while they don't if the object has been created with `Object.create(null)`.
It's probably better to iterate with `Object.keys` than with `for in` because of this, but I default to using `Object.create(null)` anyway when creating "data objects" just in case they are then iterated with a `for in` (and by planetary alignment the object prototype is also modified).

2767mr: I don't think this is relevant in the map editor since we neither modify the object prototype nor use `for in`. (As far as I remember - there might be some instances that snuck past me)

Image-unavailable: So just use `{}` for simplicity?

2767mr: Yes, that would probably better. For consistency as well

PR #241, opened 2022-08-27
Replace `hide` string property with `hidden` boolean property (issue #230)

### LINKED ISSUE #230: Bug: Wrong attribute name in NPC state editor
The editor for NPC states has an incorrect attribute name.
image

**Hide** is not the correct name for this attribute. Regardless of whether **hide** is true of false, the NPCwill remain visible. That is because the correct name for this attribute is **hidden**.

So the map editor will create a string attribute for **hide** which is _useless_. The **hidden** attribute needs to be set manually in the text editor.
image

Requesting the map editor to change **hide** to **hidden**.

Image-unavailable:
Replaces the `hide` property of NPCs with the correct `hidden` property, saved as a boolean value. (Fixes #230)


Vegita2 (changes requested):
using [`mat-select`](https://material.angular.io/components/select/overview) instead of the default `select` should solve the issue with only string values beeing set.

Vegita2 (approved):

PR #244, opened 2022-09-10
Fix collisions between bool props with same name (issue #240)

### LINKED ISSUE #240: Bug: Can't set "wait" attribute inside action steps
The DO_ACTION event step allows you to define a sequence of action steps. However in the action editor, any action step with a "wait" property will not be able to be set. Trying to set this property will set it for the DO_ACTION event, not the action step you are trying to edit.
image

Exiting the action editor after trying to click this checkbox will show that the "wait" property has been wrongly updated for the DO_ACTION event.

Some examples I have observed this with are the action steps SHOW_EFFECT and JUMP.

Image-unavailable:
Solves issues that arise if a checkbox in an action step refers to a property with the same name as one present in an event step. (Fixes #240)


Vegita2:
I think the best solution would be to remove all id's in widgets. 
The idea was to make the label clickable to select the input, but that doesn't even work half the time, so I don't think we will lose much if we just remove them altogether. This could also happen to other widgets that set an id.

Vegita2 (approved):

PR #249, opened 2022-10-22
Moved files around and removed modules

2767mr:
Moved shared files into services, components, directives and pipes.
I removed most modules because they don't really do anything except make the component declaration confusing. While it's generally a good idea to split things up, most things in the map editor are closely linked and can't really be split up without cyclic references.
Also fixed some filenames and ordered all imports.
There should be no change in functionality.


Vegita2 (approved):
Just wrote a whole paragraph about how I don't like the components folder mixing reusable components with others, only to realize that it was always like that and the pr doesn't change anything there at all.

For the future I would suggest having features split into their own folders/modules instead of separating them by type (component, directive, etc).

PR #253, opened 2022-11-17
Fix langlabel widget not resizing

Image-unavailable:


Vegita2:
what does the pull request do? Resizing behavior on `master` looks fine to me

lexisother:
If a langlabel is wrapped to have two lines due to the length, and the sidebar is resized to allow it to be on one line, it does not resize. The PR fixes this.
image

PR #258, opened 2023-03-12
Changed default vector values to 0

### LINKED ISSUE #257: Change ScalableProp default patternOffset to 0,0
When you create a new ScalableProp in the map editor, the default **patternOffset** is 1,1 instead of 0,0. This is a minor annoyance because it makes the sprite off by one pixel in each direction. Please change the default to 0,0.

image

2767mr:
Fixes #257 

Caution: Untested because it's midnight and I don't want to downgrade my nodejs to test a 2 char change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 ### Changed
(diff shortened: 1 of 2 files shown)


Vegita2:
The reason vector starts with 1 and not 0 is because when entity size is defined with a vector, it's impossible to select it again once unselected if size is 0

lexisother:
This is now handled by #271 (e56f47a5c99ac2c060c71153716b59ee24a1c75c)

PR #259, opened 2023-04-12
Fix enemyInfo

### LINKED ISSUE #256: Bug: EnemyInfo box in the event editor does not save data
Notably on eventsteps such as SPAWN_ENEMY, the event editor supports showing an EnemyInfo box, but it does not seem to actually save the data.

A video can be found [in this message](https://ptb.discord.com/channels/382339402338402315/383264360518123520/1053729411075149904).

2767mr:
Warning: Not tested because my setup cant handle old angular.
Resolves #256


Vegita2 (approved):
testing in production

PR #264, opened 2023-06-10
Open event and npc editor on double click

2767mr:
After unsuccessfully double clicking an event on instinct for the 10th time I decided to make it a feature.

### DIFF
--- a/webapp/src/app/services/phaser/entities/entity-manager.ts
+++ b/webapp/src/app/services/phaser/entities/entity-manager.ts
@@ -26,9 +26,11 @@ export class EntityManager extends BaseObject {
 	private skipEdit = false;
 	
 	private leftClickOpts: {
+		prevTimer: number;
 		timer: number;
 		pos: Point;
 	} = {
+			prevTimer: 0,
 			timer: 0,
 			pos: {x: 0, y: 0}
 		};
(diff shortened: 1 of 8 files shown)


webapp/src/app/services/phaser/entities/entity-manager.ts
```diff
@@ -177,6 +179,10 @@ export class EntityManager extends BaseObject {
 						const p = {x: pointer.worldX, y: pointer.worldY};
 						if (this.leftClickOpts.timer < 200 && Vec2.distance2(p, this.leftClickOpts.pos) < 10) {
 							this.selectEntity(entity, this.multiSelectKey.isDown);
+
+							if (!this.multiSelectKey.isDown && this.leftClickOpts.prevTimer < 500) {
+								entity.doubleClick();
```
Vegita2, line 184: can you add the proper type to entity, so it's easier for the ide to find this function usage

Vegita2 (changes requested):
code looks good. 
During testing I kept opening the event editor even though I just wanted to select the entity. there should be a check to make sure the mouse hasn't moved since the last click. 
E.g. trying to click on an entity but missing by a bit, then clicking on it for real triggers the double click

Vegita2 (changes requested):
moving an entity and clicking once still triggers the double click

https://github.com/CCDirectLink/crosscode-map-editor/assets/9483499/b629101c-83f7-40a1-82a5-1ac84d83f43e

Issue #267, opened 2023-07-15
use clipboard for entity copy


Vegita2:
copying an entity should save it to the clipboard to make it easier to use it in other instances of the editor or also allow copying entities from json files

PR #271, opened 2023-07-21
added graphical selector for some entities and fixed some issues

Vegita2:
Added New graphical selection widget for some entities:
### Prop
image

### ScalableProp
image

### Enemy
image

### ItemDestruct
image

## other changes
- Updated Phaser to 3.60.0
- Entity copy/paste now uses the clipboard to make it easier working with multiple editor instances #267
- All `Prop` Entities should now be rendered correctly
- Improved `ScalableProp` rendering. `ends` are now rendered correctly.
- Releasing mouse button outside the canvas now properly triggers. Avoids having issues like continue dragging entities even if the mouse button was released


webapp/src/app/services/phaser/entities/entity-manager.ts
```diff
@@ -355,27 +367,58 @@ export class EntityManager extends BaseObject {
...
+		const json = await navigator.clipboard.readText();
+		let entities: MapEntity[] = [];
+		try {
+			let parsed = JSON.parse(json);
+			if (!Array.isArray(parsed)) {
+				parsed = [parsed];
+			}
+			entities = (parsed as any[]).filter(v => this.isMapEntity(v));
+		} catch (e) {
+			console.warn('could not parse entities from clipboard');
```
2767mr, line 397: This should probably also show a snackbar warning

2767mr (changes requested):
Things I noticed:
* A loading indicator would be nice
* Search should include sub types (with toggle?)
* Perhaps it would look better if this is centered?
  image
* I think this could use some extra margin-right
  image
* There is no good way to see the full name of the character now. In my other projects I usually do `text-overflow: ellipsis` with a tooltip containing the full text. This would clash with the existing tooltips here but maybe we can figure something out.
  image
* Creating a new ScalableProp creates a 1x1 px entity (was already the case before this PR)
* I thought about opening the prop picker after double clicking props but that would probably be a bit much. Maybe opening it after creating a new prop would be nice?


Things I really liked:
* The "transparent texture"
* The size change on hover
* The red outline
* The counter in the top right of the props
* The prop rendering. That one has bothered me for a long time.

Vegita2:
> Perhaps it would look better if this is centered?

Is this possible? the container uses `flex-wrap: wrap;` so i'm not sure how you could center them after the wrapping

> I thought about opening the prop picker after double clicking props but that would probably be a bit much. Maybe opening it after creating a new prop would be nice?

We could always trigger the double click when adding a new entity, but not sure if it's always the first thing you want to do.  I would keep it like it currently is for now and decide later with some user feedback. It could also be toggleable in the settings menu

* [ ] A loading indicator would be nice
* [ ] Search should include sub types (with toggle?)
* [ ] I think this could use some extra margin-right
   image
 * [ ] There is no good way to see the full name of the character now. In my other projects I usually do `text-overflow: ellipsis` with a tooltip containing the full text. This would clash with the existing tooltips here but maybe we can figure something out.
   image
* [x] Creating a new ScalableProp creates a 1x1 px entity (was already the case before this PR)

lexisother:
Can this be made to scroll down to the selected prop/enemy/destructible when you open the modal again?

2767mr (approved):

Vegita2:
@lexisother made a new issue for it #277

PR #285, opened 2023-08-06
add manual input to graphical select widget

### LINKED ISSUE #278: Graphical Selection: Add support for manual entry
The new graphical selection widget for NPCs, Enemies, props, etc is very cool and useful. However, it does not allow these values to be manually specified. Modders who have custom NPCs, enemies, etc will need to manually set these in the json file now.

Ideally the widget would allow manual string entry similar to how https://github.com/CCDirectLink/crosscode-map-editor/pull/233 allows manual string entry with dropdowns.

image

Vegita2:
fixes #278

Added input so types from mods or missing ones can still be used.

image

Not added for `ScalableProp` and `ItemDestruct` because they have additional logic for setting a value

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+
+### Changed
+- Added manual input field to graphical selection widget for `Prop`, `Enemy`, `NPC`
+
 ## [1.0.0] 2023-08-04
 
 ### Added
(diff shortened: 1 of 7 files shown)


2767mr (approved):
Looks good, although I would have expected it to be above the filter or even in the same textbox

lexisother (changes requested):
I'd prefer if this PR also added support for `ScalableProp` and `ItemDestruct` before it is merged.

Vegita2:
> I'd prefer if this PR also added support for `ScalableProp` and `ItemDestruct` before it is merged.

`ItemDestruct` is supported. 
`ScalableProp` uses a complex object which can't be easily edited with a single input field. You can add a new issue for it if you want it.

Issue #115, opened 2019-09-19
Implement an character selector.


keanuplayz:
For NPCs, events, etc.

Vegita2:
what's a character selector? I need some more information
- where is it used? (Event type, Entity type)
- how should it behave

keanuplayz:
It should be used in situations like this.
![image](https://user-images.githubusercontent.com/50610755/65271318-39b1d100-db1d-11e9-94de-dbbb3ba5935a.png)

Behavior should be like a window, where you can go into a character folder (ex. bergen), and select a person. If possible, let the editor preview the person you're selecting.

Vegita2:
Preview sounds like a nice feature.
It's not easy to implement so it might take a while until we can get that to work properly.

PR #287, opened 2023-08-06
improves person-expression widget

### LINKED ISSUE #284: SHOW_MSG is guesswork to use
Now that the NPC selector has been implemented, it is also in effect on SHOW_MSG and SHOW_SIDE_MSG. Obviously not every character has a portrait, yet still everyone is listed. Preferably, a separate selector should be made for these events and additionally an expression selector would be fitting as well.

image

Vegita2:
- added graphical selection widget to `person-expression`
- event `person-expression` now only shows characters that have a face, fixes #284

image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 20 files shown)


lexisother:
Whoever's merging this, either close #115 manually or add "Closes #115" to the commit description.

2767mr (approved):

Issue #157, opened 2020-04-19
Allow object levels in layers


2767mr:
In order for `ObjectLayerView`s to work the level select in the layer settings must be extended with the `object1`, `object2` and `object3` layers.

Related to #119

PR #293, opened 2023-08-15
Added support for selecting the special layer levels

### LINKED ISSUE #286: Add support for creation of missing layer level types
Allow the creation/setting of layers of all special map levels:
- `object1`
- `object2`
- `object3`
- `last`
- `first`
- `light`
- `postlight`

Currently, the only way to set layers to this are by manually going into the JSON map data. The map editor already has support for rendering/saving these layers (albeit displaying them on "level 10"), the only thing missing is the user being able to set them.

Related to issue #157.

EL20202:
Adds in support for selecting all the special level types in a layer - specifically `first`, `last`, `light`, `postlight`, `object1`, `object2`, and `object3`. The map editor already has support for loading (and saving) layers with named levels, this merely adds the ability for one to select them (as well as properly displaying them in the layer selector). 

For example, by loading up the map `autumn/path-3-1` in the current version of the map editor, followed by the same map loaded with the new support:
image image

As for the selector for the layer levels:
image

These changes also follow the current behavior of the map editor - it treats `first` as though it is level 0, and every other level as though it was level 10.

Closes #286 and #157.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+### Added
+- Layer level selector now also supports selecting all the named layer levels: `first`, `last`, `light`, `postlight`, `object1`, `object2`, and `object3`.
+
+### Changed
+- Named layer levels will now appear in the layer selector as opposed to a layer appearing on level 0 or 10.
+
+### Fixed
+- Layers on the `first` level should now properly render as though they are below level 0.
+
 ## [1.0.0] 2023-08-04
 
 ### Added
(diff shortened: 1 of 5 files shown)


Vegita2:
Is it possible to make `first` be always below level 0? I noticed it in a few maps that first is sometimes drawn on top of another layer on level 0

EL20202:
I'll see what I can do, I have some ideas on how that could be accomplished.

EL20202:
There, I think that should do it - from my quick tests with `jungle/path-01-entrance.json` (which has a layer at `first`), I didn't have any issues, especially with going into 3D, which is where I expected to have issues.

EL20202:
There, I added the changelog. Everything should be ready now.

Vegita2 (changes requested):
good addition.

I thought level -1 would cause some issues, but I couldn't find any problems with it.

EL20202:
There, made the requested changes. And yeah - I too was a little surprised too how level -1 didn't break much of anything - if anything at all.

2767mr (approved):

Vegita2 (approved):

PR #300, opened 2023-08-25
added cache for loading images/json

Vegita2:
avoids loading the same images/json multiple times in parallel.

This should improve performance for graphical selection widget a little bit

### DIFF
--- a/webapp/src/app/services/phaser/helper.ts
+++ b/webapp/src/app/services/phaser/helper.ts
@@ -8,6 +8,11 @@ import Scene = Phaser.Scene;
 
 export class Helper {
 	
+	// stores promises so editor doesn't try to load the same file multiple times simultaneously
+	private static cache: {
+		[key: string]: Promise<any> | undefined;
+	} = {};
+	
 	public static worldToTile(x: number, y: number): Point {
 		const p: Point = {x: 0, y: 0};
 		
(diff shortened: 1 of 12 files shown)


webapp/src/app/services/phaser/helper.ts
```diff
@@ -72,31 +77,33 @@ export class Helper {
...
 		
-		// get json from cache
-		if (scene.cache.json.has(key)) {
-			return callback(scene.cache.json.get(key));
+		if (scene.cache.json.has(jsonKey)) {
+			return scene.cache.json.get(jsonKey);
+		}
+		
+		if (this.cache[jsonKey]) {
+			return this.cache[jsonKey];
```
2767mr, line 89: At this point, we might as well do everything ourselves. We just need to replace the instances of `json.get(` with the helper function.
```ts
const jsonKey = key + '.json';

if (this.cache[jsonKey]) {
	return this.cache[jsonKey];
}

const promise: Promise<T | undefined> = firstValueFrom(
	Globals.httpService.resolveFile(jsonKey)
		.pipe(
			map(file => JSON.parse(file)),
			catchError(() => {
				console.warn(`Failed to resolve resource: ${jsonKey}`);
				return of(undefined);
			})
		)
	);

this.cache[jsonKey] = promise;
return promise;
```

2767mr (changes requested):

lexisother:
This breaks the tile selector for some reason.

webapp/src/app/services/phaser/helper.ts
```diff
@@ -115,15 +122,23 @@ export class Helper {
 			return true;
 		}
 		
-		// TODO: save promise to avoid loading the same texture simultaneously
+		if (this.cache[key]) {
+			return this.cache[key];
+		}
+		const promise = this.loadTextureInternal(key, scene);
+		this.cache[key] = promise;
+		return promise;
```
2767mr, line 130: I completely missed this code. This doesn't work because the cache has to be dependant on the scene that is passed in as an argument.

2767mr (changes requested):

Vegita2:
Closing this because I don't think these changes make too much of a difference and I don't want to fix the remaining issues

PR #303, opened 2023-09-02
Added confirm close dialog

### LINKED ISSUE #302: Loading a new map should show a warning about losing unsaved changes
If the user has unsaved changes they should be shown a warning when loading a new map or closing the editor.

2767mr:
Resolves #302 
image

### DIFF
--- a/webapp/src/app/app.component.ts
+++ b/webapp/src/app/app.component.ts
@@ -25,7 +25,9 @@ export class AppComponent {
 		if(this.eventsService.hasUnsavedChanges.getValue()) {
 			$event.returnValue = 'Are you sure you want to discard your changes?';
 			
-			const dialogRef = this.overlayService.open(ConfirmCloseComponent);
+			const dialogRef = this.overlayService.open(ConfirmCloseComponent, {
+				hasBackdrop: true,
+			});
 			dialogRef.instance.showDevMode = true;
 			dialogRef.ref.onClose.subscribe(result => {
 				if (result) {
--- a/webapp/src/app/app.component.ts
+++ b/webapp/src/app/app.component.ts
@@ -1,4 +1,8 @@
-import {Component} from '@angular/core';
+import { Component, HostListener } from '@angular/core';
+import { MatDialog } from '@angular/material/dialog';
+import { Router } from '@angular/router';
+import { ConfirmCloseComponent } from './components/dialogs/confirm-close/confirm-close.component';
+import { GlobalEventsService } from './services/global-events.service';
 
 @Component({
 	selector: 'app-root',
@@ -6,5 +10,31 @@ import {Component} from '@angular/core';
 	styleUrls: ['./app.component.scss']
 })
 export class AppComponent {
+	constructor(
+		private readonly eventsService: GlobalEventsService,
+		private readonly dialogService: MatDialog,
+		private readonly router: Router
+	) { 
+		this.router.events.subscribe(event => {
+			console.log(event.constructor.name, event);
+		});
+	}
 
+	@HostListener('window:beforeunload', ['$event'])
+	doSomething($event: any) {
+		if(this.eventsService.hasUnsavedChanges.getValue()) {
+			$event.returnValue = 'Are you sure you want to discard your changes?';
+			
+			const dialogRef = this.dialogService.open(ConfirmCloseComponent);
+			dialogRef.afterClosed().subscribe(result => {
+				if (result) {
+					this.eventsService.hasUnsavedChanges.next(false);
+					window.close();
+				} else if (result === null) {
+					this.eventsService.hasUnsavedChanges.next(false);
+					location.reload();
+				}
+			});
+		}
+	}
 }
--- a/webapp/src/app/app.component.ts
+++ b/webapp/src/app/app.component.ts
@@ -21,7 +21,7 @@ export class AppComponent {
 	}
 
 	@HostListener('window:beforeunload', ['$event'])
-	doSomething($event: any) {
+	onUnload($event: any) {
 		if(this.eventsService.hasUnsavedChanges.getValue()) {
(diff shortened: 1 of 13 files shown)


lexisother:
I haven't pulled this in locally but if the design is as shown on the image I'm not sure if that really fits in. It looks kludgy.

Something like this would be better. Bigger, bold, left-aligned centered header text; smaller, non-bold description text.

image

2767mr:
image

webapp/src/app/app.component.ts
```diff
@@ -1,10 +1,41 @@
...
+				if (result) {
+					this.eventsService.hasUnsavedChanges.next(false);
+					window.close();
+				} else if (result === null) {
+					this.eventsService.hasUnsavedChanges.next(false);
+					location.reload();
+				}
+			});
+		}
+	}
```
Vegita2, line 40: I would prefer to use angular guards instead of using window events directly, https://angular.io/api/router/CanDeactivateFn

Vegita2, line 40: I would prefer to use angular guards instead of using window events directly, https://angular.io/api/router/CanDeactivateFn

webapp/src/app/components/dialogs/confirm-close/confirm-close.component.html
```diff
@@ -0,0 +1,16 @@
...
+  >
+	<ng-container ngProjectAs="content">
+    <div class="p-3 text-center">
+      Are you sure you want to discard your changes?
+    </div>
+	</ng-container>
+	<ng-container ngProjectAs="actions">
+    <button mat-stroked-button (click)="close(false)">Go back</button>
+    <button mat-stroked-button (click)="close(null)" *ngIf="showDevMode && devMode" cdkFocusInitial>DEV: Reload</button>
+    <button mat-stroked-button (click)="close(true)" cdkFocusInitial>Discard changes</button>
```
Vegita2, line 14: I think it would be better to have the initial focus be on "Go back". It's not clear if the user really wanted to discard everything or just forgot to save, so we should default on the harmless option

webapp/src/app/app.component.ts
Vegita2, line 40: I would prefer to use angular guards instead of using window events directly, https://angular.io/api/router/CanDeactivateFn

Vegita2, line 40: I would prefer to use angular guards instead of using window events directly, https://angular.io/api/router/CanDeactivateFn

webapp/src/app/components/dialogs/load-map/load-map.component.ts
```diff
@@ -67,13 +73,31 @@ export class LoadMapComponent {
 		this.mapsSource.data = this.virtualRoot.children || [];
 		this.ref.detectChanges();
 	}
+
+	private async showConfirmDialog() {
+		const hasUnsavedChanges = await firstValueFrom(this.eventsService.hasUnsavedChanges);
```
Vegita2, line 78: you used `this.eventsService.hasUnsavedChanges.getValue()` in previous code. Is there a reason why it's different here?

webapp/src/app/external-modules/material.module.ts
```diff
@@ -45,6 +46,7 @@ const MODULES = [
 	MatTreeModule,
 	MatDividerModule,
 	MatSlideToggleModule,
+	A11yModule,
```
Vegita2, line 49: Is this used anywhere?

Vegita2 (changes requested):
Looks good, just one thing that bothers me. The Dialog should be centered. Currently the top left edge is in the center
image

Also, what is the purpose of dev mode?

webapp/src/app/app.component.ts
2767mr: `CanDeactivate` doesn't actually prevent the window from closing.

webapp/src/app/components/dialogs/load-map/load-map.component.ts
2767mr: onUnload cannot be async so I had to use `getValue` which I consider bad practice

webapp/src/app/external-modules/material.module.ts
2767mr: Yes, that enables the `cdkFocusInitial` directive

2767mr:
> Also, what is the purpose of dev mode?

If we are developing and make a change the message will pop up. I want to be able to press enter in that case to reload the window. Unfortunately we don't know if it's a reload or a window close so we need different buttons.

Vegita2:
> If we are developing and make a change the message will pop up. I want to be able to press enter in that case to reload the window. Unfortunately we don't know if it's a reload or a window close so we need different buttons.

In that case we could completely disable the confirmation dialog in dev mode. I expect it to be quite annoying when developing

2767mr:
I thought about that but most of our users are powerusers. I would expect EL or Alyxia to work in dev mode. Maybe an environment variable would be a compromise? I also made sure to have it focused on popup so you only have to press enter which should be next to no overhead.

EL20202:
If I may interject - why not make that a normal setting? Make it so people can turn off the save confirmation regardless of it being in dev mode or not. It can default to being on, and if someone is doing something where they're inconvenienced by the save window for whatever reason, they can just turn it off (at their own risk, of course)

2767mr:
I have never seen any program where you can turn it off and I don't really see someone getting annoyed at it.

Vegita2:
> I would expect EL or Alyxia to work in dev mode.

dev mode reduces performance. I don't see why anyone would use it when working on maps.

lexisother:
Use of features that take long to merge? Also, the production builds for Linux are made with a version of glibc that my computer doesn't like. Phaser refuses to load for me without building locally.

Vegita2 (approved):

PR #320, opened 2024-07-05
Fix window resizing

### LINKED ISSUE #319: Resizing the window now breaks the game canvas
As of version 1.6.0, I am seeing the canvas resolution break when resizing the window. This makes opening a second session and moving it to another monitor very difficult. Restarting the map editor will fix the issue.

I suspect that the fix for https://github.com/CCDirectLink/crosscode-map-editor/issues/265 must have caused this new issue.

I am using Windows 10 with multiple 1920x1080 monitors.

map-editor-resolution-issue

2767mr:
Hopefully closes #319

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+
+### Fixed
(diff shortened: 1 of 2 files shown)


Vegita2:
I didn't try it yet, but do you know why this solves it? I never really understood how that zoom property works and came to the current version through trial and error

2767mr:
When I was debugging it, it seemed to be a Phaser bug: If you call `setZoom` it believes that the size of the canvas has been adjusted as well (even though it hasn't) and then no longer resizes the canvas upon calling `resize` since it believes it is already the correct size.
Not sure how the zoom property works either.

PR #324, opened 2024-07-20
Support ccmod.json and mod names

### LINKED ISSUE #323: Support ccmod.json
`ccmod.json` is the replacement for `package.json`. New mods likely only have the former. The format is very similar so this should be easy to implement. We just need to do it.

2767mr:
Closes #323 and also implements proper mod names instead of just directory names


2767mr:
image

Vegita2 (approved):

Vegita2 (approved):

PR #348, opened 2026-04-03
moved "vanilla maps" checkbox to map loading

### LINKED ISSUE #347: Improve load map dialog by adding dropdown to swap between modded or vanilla maps
Currently, the map editor when having a mod loaded only shows maps from that mod unless one has the "Include vanilla maps" option selected. However, one clear downside to that option is that the load map dialog makes zero distinction between a modded and a vanilla map. Along with that, there's no way to enable the display of vanilla maps while having a mod loaded without restarting the map editor. 

To rectify all of the above issues, I propose the addition of a dropdown menu in the load map dialog which will allow one to switch between displaying modded maps and vanilla maps. A mockup is displayed below.

Image

The dropdown would contain the options of "Vanilla maps" (or just "Vanilla" if preferred) as well as, for example, "Mod Name" to switch between displaying maps just from vanilla or from a loaded mod (by default, it likely should display maps from the loaded mod instead as that is the default behavior of the map editor currently.)

The reason I propose a dropdown rather than, say, a checkbox to toggle displaying vanilla maps is that this offers much greater expandability. For example, if support was added for loading maps from extensions (such as the postgame) it could be done by adding more options to the dropdown, though adding support for extensions goes beyond the scope of this issue.

I would also propose that with the addition of this dropdown, the option to "include vanilla maps" in the settings menu likely should be removed, as it would be obsolete by the addition of a dropdown to swap between modded/vanilla maps on the fly.

Vegita2:
Moved the "vanilla maps" checkbox from the settings page to the map loading dialog. It's now possible to load either maps from the selected mod or vanilla maps. Fixes #347 

image

### DIFF
--- a/webapp/src/app/components/dialogs/load-map/load-map.component.ts
+++ b/webapp/src/app/components/dialogs/load-map/load-map.component.ts
@@ -19,7 +19,7 @@ import { MatFormField, MatInput } from '@angular/material/input';
 import { FormsModule } from '@angular/forms';
 import { HighlightDirective } from '../../../directives/highlight.directive';
 import { MatProgressSpinner } from '@angular/material/progress-spinner';
-import { SettingsService } from '../../../services/settings.service';
+import { SHARED_SERVICE } from '../../../services/shared-service';
 import { MatCheckbox } from '@angular/material/checkbox';
 
 
@@ -54,7 +54,7 @@ export class LoadMapComponent {
 	private searchFilterService = inject(SearchFilterService);
 	private readonly eventsService = inject(GlobalEventsService);
 	private readonly overlayService = inject(OverlayService);
-	private readonly settingsService = inject(SettingsService);
+	private readonly sharedService = inject(SHARED_SERVICE);
 	
 	
 	@ViewChild('fileUpload', {static: true})
@@ -85,7 +85,7 @@ export class LoadMapComponent {
 		});
 		
 		this.mapsSource.data = [];
-		this.currentMod = this.settingsService.sharedService.getSelectedMod();
+		this.currentMod = this.sharedService.getSelectedMod();
 	}
 	
 	focusInput() {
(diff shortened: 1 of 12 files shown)


webapp/src/app/components/dialogs/load-map/load-map.component.ts
```diff
@@ -71,22 +75,31 @@ export class LoadMapComponent {
...
 	filter = '';
 	
+	currentMod = '';
+	vanillaMaps = signal(false);
+	
 	constructor() {
+		effect(() => {
+			this.vanillaMaps();
+			this.refresh();
+		});
```
2767mr: For situations like these I would highly recommend using [resources](https://angular.dev/guide/signals/resource) or at least use `untracked(() => this.refresh())` to avoid unexpected refreshes.

webapp/src/app/components/dialogs/load-map/load-map.component.ts
```diff
@@ -71,22 +75,31 @@ export class LoadMapComponent {
...
-		this.http.getMaps().subscribe(paths => {
-			this.loading = false;
-			this.displayMaps(paths);
-			this.update();
-		});
+	async refresh() {
+		const req = this.vanillaMaps() ? this.http.getVanillaMaps() : this.http.getMaps();
+		this.loading.set(true);
+		const paths = await firstValueFrom(req);
+		this.loading.set(false);
```
2767mr: I prefer loading to be set to false even if the request fails

webapp/src/app/services/settings.service.ts
```diff
@@ -2,18 +2,16 @@ import { Injectable } from '@angular/core';
...
 
 @Injectable({
 	providedIn: 'root'
 })
 export class SettingsService {
-	
+
 	private settings: AppSettings = {
 		wrapEventEditorLines: this.loadBooleanOrDefault('wrapEventEditorLines', true),
-		includeVanillaMaps: this.loadBooleanOrDefault('includeVanillaMaps', false),
```
2767mr, line 16: Wouldn't it make sene to keep the setting, even though it is no longer in the settings dialog? I.e. ideally it would always be updated when the user changes the checkbox.

2767mr (approved):
Looks good, just a few nitpicks

webapp/src/app/components/dialogs/load-map/load-map.component.ts
Vegita2: Good point, should use resource

webapp/src/app/services/settings.service.ts
Vegita2: Also good point, I'll change it

2767mr (approved):

PR #354, opened 2026-05-24
Character selector previews now render NPCs facing south.

Vegita2:


2767mr (approved):
For some reason this changes this changes the opacity of the character images on the right side of the dialog.
Before:

Image

After:

Image

I doubt it makes a difference though.

PR #357, opened 2026-09-13
Preserve map entity order

### LINKED ISSUE #218: Save data in the same order as loaded
Currently when loading a map, certain entries such as entities are reordered.
One should be able to load a file and save it without any change to the file.
This improves consistency and helps automated tools such as diff.

2767mr:
Since entities are generated concurrently, the order at which they are added was indeterminstic. Instead, this waits for all entities to load and then adds all successfully generated entities in the correct order.

This has the downside of not adding any entities if any entity never finishes loading.

I also included a fix for a race condition while I was at it.


webapp/src/app/services/phaser/entities/entity-manager.ts
```diff
@@ -314,7 +314,15 @@ export class EntityManager extends BaseObject {
...
-		await Promise.all(promises);
+
+		const newEntities = await Promise.allSettled(promises);
+		for (const newEntity of newEntities) {
+			if (newEntity.status === 'fulfilled') {
+				this._entities.push(newEntity.value);
+			} else {
+				console.error('Error while generating entity: ', newEntity.reason);
+			}
+		}
```
Vegita2: I'd prefer it if it would still use `Promise.all` to make sure that maps with errors don't load. Because right now it would silently drop entities that couldn't be loaded and load/save a map would drop that entity entirely

Vegita2 (changes requested):
looks good. I remembered having a test case fail because of entity orders and found this line
```
// TODO: fix map loading, order of entities doesn't matter
```
Now it probably should say it DOES matter. Tests currently don't work anyway so it's something for the future when we update angular and use vitest.

I also noticed there is a race condition when quickly loading multiple maps which already existed before. Because of the async entity loading the second map can load entities from the first map. Probably out of scope for this pull request but should be kept in mind

webapp/src/app/services/phaser/entities/entity-manager.ts
2767mr: Done

Vegita2 (approved):

